PR #16, opened 2025-07-21
Session Context

LuckyIYI:
A mechanism to cache prefilled text chunks as and object to allow their reuse in following requests


cli/src/server/chat_completions.rs
```diff
@@ -12,59 +14,91 @@ use uzu::session::{
...
+    let model_name = state.model_name.clone();
+    let system_prompt_key = request.system_prompt_key;
+    let tokens_limit = request.max_tokens.unwrap_or(2048);
+    let messages: Vec<SessionMessage> = request
+        .messages
+        .into_iter()
+        .map(|m| SessionMessage {
+            content: m.content,
+            role: match m.role.as_str() {
+                "system" => SessionMessageRole::System,
```
eugenebokhan: can `SessionMessageRole` be passed from outside instead of matching?

uzu/src/backends/metal/kernel/attention/mod.rs
```diff
@@ -575,7 +575,10 @@ impl EncodableWithState for AttentionKernelEncodable {
             )
         };
 
-        let prefix_length = state.kv_cache.borrow().data[self.layer_index].projected_effective_prefix_length(parameters.projection_step.unwrap_or(0));
+        let prefix_length = state.kv_cache.borrow().data[self.layer_index]
+            .projected_effective_prefix_length(
+                parameters.projection_step.unwrap_or(0),
```
eugenebokhan, line 580: What will happen on `0`?

uzu/src/session/session.rs
```diff
@@ -127,6 +140,39 @@ impl Session {
     where
         F: Fn(SessionOutput) -> bool,
     {
+        let generator = self.generator.as_mut().unwrap();
```
eugenebokhan: Maybe we should shift to throwing errors instead of endless unwraps?

eugenebokhan (commented):
Let's sort out minor comments

uzu/src/backends/metal/kernel/attention/mod.rs
LuckyIYI: it's old code, just formatted 
if 0, it just won't do pre-encoding (async encoding), it's harmless

PR #36, opened 2025-10-13
New speculator api

uuuvn:


crates/uzu/src/linearizer/trie.rs
```diff
@@ -20,6 +36,57 @@ impl TokenTrie {
...
+            current_node =
+                current_node.get_or_insert_next(*prefix.last().unwrap());
+            trie_size += 1;
+        }
+
+        let mut speculation_prefix = prefix.to_vec();
+        while trie_size < max_length {
+            let speculation_result = speculator.speculate(&speculation_prefix);
+
+            let mut arr =
```
norpadon, line 60: No need for let mut, also arr is a bad name

crates/uzu/src/linearizer/trie.rs
```diff
@@ -20,6 +36,57 @@ impl TokenTrie {
...
+                speculation_result.into_iter().collect::<Vec<(u64, f32)>>();
+            arr.sort_by(|(_, a), (_, b)| f32::total_cmp(b, a));
+
+            arr = arr
```
norpadon: Let arr

crates/uzu/src/linearizer/trie.rs
uuuvn: sort can only be done on a mutable vec, there is no sort on iterator or on non-mutable vec

PR #41, opened 2025-11-11
policy/sampling separation

uuuvn:


crates/uzu/src/backends/metal/kernel/processing/gumbel.metal
```diff
@@ -0,0 +1,43 @@
+#include <metal_stdlib>
+#include "../definitions.metal"
+#include "../rng.metal"
+
+#define BLOCK_SIZE 64
```
LuckyIYI: we usually define BLOCK_SIZE as number of threads per threadgroup, better to keep it consistent

the inner loop probably should be called CHUNK/GRAIN _SIZE, or something more suitable for context

PR #50, opened 2025-11-25
Trie rewrite and guuumbel speculative decoding

uuuvn:
Guuumbel speculator trick: both speculator and llm sample via gumbel max trick using the same noise for increased acceptance rate


crates/uzu/src/generator/generator.rs
```diff
@@ -72,80 +72,70 @@ impl Generator {
...
+        let flat_trie = suffix_root.linearize();
+        let active_suffix_length = flat_trie.len() - 1;
+
+        let token_ids = tokens
+            .iter()
+            .copied()
+            .take(tokens_length - 1)
+            .chain(flat_trie.token_ids())
+            .chain(repeat_n(0, suffix_length - active_suffix_length))
+            .collect::<Vec<u64>>();
```
norpadon: Why collect to Vec when you can collect to Box<[...]>?

norpadon (approved):
Lgtm

PR #51, opened 2025-11-27
XGrammar policy

uuuvn:
Depends on #50

Built on top of @eugenebokhan's `xgrammar` branch

### DIFF
--- a/crates/uzu/src/backends/metal/kernel/sampling/bitmask.metal
+++ b/crates/uzu/src/backends/metal/kernel/sampling/bitmask.metal
@@ -1,0 +1,58 @@
+#include <metal_stdlib>
+#include "../definitions.metal"
+
+#define GRAIN_SIZE 64
+#define BLOCK_SIZE 1024
+
+#define BITS_IN_U32 32
+
+template <typename T>
+void batched_bitmask(
+    device const T* logits_data,
+    device const uint32_t* bitmask_data,
+    device T* processed_logits,
+    constant uint& vocab_size,
+    uint group_idx,
+    uint batch_idx,
+    uint thread_idx
+) {
+    uint base_idx = batch_idx * vocab_size + group_idx * BLOCK_SIZE * GRAIN_SIZE + thread_idx;
+    uint batch_end = batch_idx * vocab_size + vocab_size;
+
+    #pragma unroll(4)
+    for (uint i = 0; i < GRAIN_SIZE; i++) {
+        uint global_idx = base_idx + i * BLOCK_SIZE;
+        if (global_idx < batch_end) {
+            bool mask = (bitmask_data[global_idx / BITS_IN_U32] >> (global_idx % BITS_IN_U32)) & 0b1;
+            processed_logits[global_idx] = select(T(-INFINITY), logits_data[global_idx], mask);
+        }
+    }
+}
+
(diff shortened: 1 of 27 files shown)


crates/uzu/src/backends/metal/kernel/sampling/bitmask.metal
```diff
@@ -0,0 +1,56 @@
...
+    uint thread_idx
+) {
+    uint base_idx = batch_idx * vocab_size + group_idx * BLOCK_SIZE * GRAIN_SIZE + thread_idx;
+    uint batch_end = batch_idx * vocab_size + vocab_size;
+
+    #pragma unroll(4)
+    for (uint i = 0; i < GRAIN_SIZE; i++) {
+        uint global_idx = base_idx + i * BLOCK_SIZE;
+        if (global_idx < batch_end) {
+            bool mask = (bitmask_data[global_idx / 32] >> (global_idx % 32)) & 0b1;
```
eugenebokhan: what is 32? let's introduce a meaningful var / name for it

crates/uzu/src/backends/metal/kernel/sampling/mod.rs
```diff
@@ -58,6 +58,8 @@ enum ArgmaxImplementation {
 }
 
 pub struct SamplingKernel {
+    bitmask_pipeline: MTLComputePipelineState,
```
eugenebokhan, line 61: I prefer to have one pipeline state per managing object, and a higher level object simply orchestrates the lower level ones. Can we introduce `BitmaskKernel`, `TemperatureKernel`, `TopKKernel`,  so `SamplingKernel` could manage them?
Let's use the encoder pattern in our cpu code like this one in [metal-tools](https://github.com/computer-graphics-tools/metal-tools/blob/main/Sources/MetalComputeTools/Kernels/TextureCopy/TextureCopy.swift), which is quite simple: the object owns a pipeline state and has an encode function, and can encode it using encoder & in pipeline state.

crates/uzu/src/backends/metal/kernel/sampling/mod.rs
```diff
@@ -366,6 +399,38 @@ impl SamplingKernel {
...
+            3,
+            size_of::<u32>() as u64,
+            &vocab_size as *const u32 as *const std::ffi::c_void,
+        );
+
+        let elements_in_group = BLOCK_SIZE * ELEMENTWISE_GRAIN_SIZE;
+        let groups = (vocab_size + (elements_in_group as u32 - 1))
+            / elements_in_group as u32;
+
+        compute_encoder.dispatch_thread_groups(
```
eugenebokhan, line 426: Are you sure you want to use `dispatch_thread_groups` instead over `dispatch_threads`? We are targeting modern GPUs

crates/uzu/src/generator/grammar.rs
```diff
@@ -0,0 +1,203 @@
...
+            Self::Always => (),
+            Self::Triggered {
+                trigger_token_id: _,
+                trigger_distance,
+            } => *trigger_distance = None,
+        }
+    }
+}
+
+pub struct CompiledGrammar {
```
eugenebokhan, line 78: let's put each type into a separate rs file for better code navigation

PR #277, opened 2026-03-20
RHT

eugenebokhan:


crates/uzu/src/backends/metal/kernel/quant_matmul/qmm_transposed.metal
```diff
@@ -13,13 +14,15 @@ PUBLIC KERNEL(QuantizedMatmulQmmTransposed)(
     const device T* biases OPTIONAL(use_mlx_quant),
     const device T* input,
     device T* output,
+    const device int32_t* hadamard_factors OPTIONAL(use_hadamard),
```
CC-Yeh, line 17: Are values of these factors only {-1, 1}? if so, can we use `int8_t` to save bandwidth and memory?

crates/uzu/src/backends/metal/kernel/hadamard_transform/hadamard_transform.h
```diff
@@ -0,0 +1,30 @@
+#include <metal_stdlib>
+#include "../common/defines.h"
+#include "../common/thread_context.h"
+
+using namespace metal;
+
+static inline float simdgroup_hadamard_transform(
```
CC-Yeh:
```suggestion
static METAL_FUNC float simdgroup_hadamard_transform(
```

crates/uzu/src/backends/metal/kernel/quant_matmul/qmm_transposed.metal
```diff
@@ -103,4 +106,24 @@ PUBLIC KERNEL(QuantizedMatmulQmmTransposed)(
...
     }
   }
+
+  if (use_hadamard) {
+    threadgroup_barrier(mem_flags::mem_device);
+
+    const uint col = out_block_idx * 32 + simd_lane;
+    const int32_t factor = hadamard_factors[col];
+
+    for (uint r = simd_group * 8; r < simd_group * 8 + 8; r++) {
```
CC-Yeh: magic number `8` is a bit confusing.

Can we have sth like:

```c++
 constexpr uint rows_per_simdgroup = 8; // BM / num_simdgroups
```

crates/uzu/src/backends/common/kernel/quant_matmul.rs
```diff
@@ -117,45 +126,47 @@ impl<B: Backend> QuantizedMatmulKernelEncodable<B> {
...
+                MatrixMatrixKernel::QmmTransposed64x64(
+                    <B::Kernels as Kernels>::QuantizedMatmulQmmTransposed64x64Kernel::new(
+                        context,
+                        configuration.data_type,
+                        group_size,
+                        bits,
+                        use_zero_points,
+                        use_mlx_quant,
+                    )
+                    .map_err(QuantizedMatmulError::BackendError)?,
```
CC-Yeh, line 140: Are we going to support Qmm64x64(supposed it's faster for larger matrices) kernels for hadamard_transform as well?

crates/uzu/src/backends/metal/kernel/quant_matmul/qmm_transposed.metal
```diff
@@ -103,4 +106,24 @@ PUBLIC KERNEL(QuantizedMatmulQmmTransposed)(
       );
     }
   }
+
+  if (use_hadamard) {
+    threadgroup_barrier(mem_flags::mem_device);
```
CC-Yeh, line 111: How costly is synching with device memory?

Is it possible to just process outputs stored in threadgroup memory? Possibly need to change `qmm_transposed_impl`.

crates/uzu/src/backends/metal/kernel/quant_matmul/qmm_transposed.metal
uuuvn: Yeah, that requires lalamo side changes to the code that as far as I know still lives in a bunch of notebooks... But this is probably the most obvious thing to change when we have the lalamo side code and can play around with how things are stored. Maybe even don't store it at all, just re-derive from seed. It is random after all.

crates/uzu/src/backends/common/kernel/quant_matmul.rs
uuuvn: Yeah, eventually all qmv/qmm variants should support hadamard transform

crates/uzu/src/backends/metal/kernel/quant_matmul/qmm_transposed.metal
uuuvn: Pretty expensive. One of the things to optimize in followup prs

uuuvn: In general the best case scenario for us is kernels that naturally have the 32 consecutive elements in 32 lanes of one simdgroup, eg I've refactored RMSNorm to be like that to have very cheap output hadamard and it became faster on non-hadamard too. It's more complicated in cases like matmuls where it might not be the optimal way to do things on non-hadamard which is why I left it this way for now

crates/uzu/src/backends/metal/kernel/hadamard_transform/hadamard_transform.h
uuuvn: fixed

crates/uzu/src/backends/metal/kernel/quant_matmul/qmm_transposed.metal
uuuvn: fixed

CC-Yeh (approved):

PR #330, opened 2026-04-08
faster qmv

CC-Yeh:
`air.convert.f.f32.s.i32` (SFU, ~4c) is the latency bottleneck at M≥3. Trade it for underutilized Integer/FP units:

  ```
  Before:  load i16 → and mask → air.convert i32→f32 (SFU ~4c) → fmul → fadd
  After:   load i16 → and mask → or 0x4B000000 (~1c) → bitcast (free) → fsub (~1c) → fmul → fadd
  ```

  - Replace `qdot_zero_point` with `qdot(w, x, scale, -scale*zp, sum)` — reuses `uint4_to_fp4` bit-trick path
  - Vectorize `uint4_to_fp4` with explicit float/half/bfloat specializations
  - 80% fewer `air.convert` ops in AIR

Speedup:
  | Config           | M=2       | M=3       | M=4       |
  |------------------|-----------|-----------|-----------|
  | MLX BF16 gs128   | ~noise    | 5–12%     | 4–9%      |
  | ZP F16 gs64      | 1–12%     | 12–14%    | 12–15%    |
  
M=1 unchanged (bandwidth-bound). At M≥3 the kernel becomes compute-bound — but occupancy stays ~26% because Shader Launch hardware saturates at 54%, so extra TGs queue instead of overlapping.

### DIFF
--- a/crates/uzu/src/backends/metal/kernel/quant_matmul/quant_matmul.h
+++ b/crates/uzu/src/backends/metal/kernel/quant_matmul/quant_matmul.h
@@ -14,26 +14,26 @@ METAL_FUNC U uint_to_fp(uint32_t x) {
 }
 
 template <>
-inline bfloat uint_to_fp<bfloat>(uint32_t x) {
+METAL_FUNC bfloat uint_to_fp<bfloat>(uint32_t x) {
   return as_type<bfloat>(uint16_t(x | 0x4300u)) - bfloat(128.0f);
 }
 
 template <typename U>
-inline vec<U, 4> uint4_to_fp4(uint4 n);
+METAL_FUNC vec<U, 4> uint4_to_fp4(uint4 n);
 
 template <>
-inline float4 uint4_to_fp4<float>(uint4 n) {
+METAL_FUNC float4 uint4_to_fp4<float>(uint4 n) {
   n &= uint4(0xFu);
   return as_type<float4>(n | uint4(0x4B000000u)) - float4(8388608.0f);
 }
 
 template <>
-inline half4 uint4_to_fp4<half>(uint4 n) {
+METAL_FUNC half4 uint4_to_fp4<half>(uint4 n) {
   return half4(uint4_to_fp4<float>(n));
 }
 
 template <>
-inline bfloat4 uint4_to_fp4<bfloat>(uint4 n) {
+METAL_FUNC bfloat4 uint4_to_fp4<bfloat>(uint4 n) {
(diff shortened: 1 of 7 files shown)


uuuvn (approved):

crates/uzu/src/backends/metal/kernel/quant_matmul/quant_matmul.h
```diff
@@ -7,6 +7,37 @@ using namespace metal;
 
 #include "mma.h"
 
+// Avoids air.convert (SFU) for int → float which is slower on Apple GPU.
+template <typename U>
+inline U uint_to_fp(uint32_t x) {
```
eugenebokhan:
```suggestion
METAL_FUNC U uint_to_fp(uint32_t x) {
```

crates/uzu/src/backends/metal/kernel/quant_matmul/quant_matmul.h
```diff
@@ -187,23 +148,25 @@ inline U qdot_safe(
 
     int full = N / 4;
     for (int i = 0; i < full; i++) {
-      accum +=
-          (x_thread[4 * i] * (ws[i] & 0x000f) +
-           x_thread[4 * i + 1] * (ws[i] & 0x00f0) +
-           x_thread[4 * i + 2] * (ws[i] & 0x0f00) +
-           x_thread[4 * i + 3] * (ws[i] & 0xf000));
+      auto wi = ws[i];
+      float wf0 = uint_to_fp<float>(wi & 0xfu);
```
eugenebokhan: in some places we have auto, in others we use explicit type, let's move to the explicit declarations

crates/uzu/src/backends/metal/kernel/quant_matmul/quant_matmul.h
eugenebokhan: @CC-Yeh  this needs to be extrapolated to all the funcs

crates/uzu/src/backends/metal/kernel/quant_matmul/quant_matmul.h
CC-Yeh: removed thx

eugenebokhan (approved):

eugenebokhan:
Would be cool to see the end-to-end perf gain done via performance-benchmarks

CC-Yeh:
> Would be cool to see the end-to-end perf gain done via performance-benchmarks

Is there task for small batch (M <= 4)/speculative case?

PR #334, opened 2026-04-09
Wasi build

agolokoz:


.github/workflows/tests.yml
```diff
@@ -77,6 +77,12 @@ jobs:
             ${{ matrix.backend != 'cpu' && format('--features {0}', matrix.backend) || '' }} \
             ${{ matrix.tracing && '--features tracing' || '' }} \
 
+      - name: Build uzu for target wasi
+        if: ${{ matrix.backend == 'cpu' && !matrix.tracing }}
+        run: |
+          cargo +stable build --package uzu --no-default-features \
+            --target wasm32-wasip1-threads \
```
uuuvn: This (target) should be another matrix dimension

crates/uzu/build/main.rs
```diff
@@ -26,13 +26,18 @@ async fn main() -> anyhow::Result<()> {
...
     println!("cargo::rustc-check-cfg=cfg(metal_backend)");
     if metal_backend {
         println!("cargo::rustc-cfg=metal_backend");
     }
 
+    let grammar_xgram = cfg!(feature = "grammar") && env::var("CARGO_CFG_TARGET_ARCH")? != "wasm32";
+    println!("cargo::rustc-check-cfg=cfg(grammar_xgram)");
+    if grammar_xgram {
+        println!("cargo::rustc-cfg=grammar_xgram");
+    }
```
uuuvn: xgram? Let's do it `grammar_xgrammar`

crates/uzu/src/language_model/language_model_generator.rs
```diff
@@ -78,15 +78,15 @@ pub trait LanguageModelGeneratorTrait {
     fn prefill(
         &mut self,
         tokens: Vec<u64>,
-        compiled_grammar: Option<&mut CompiledGrammar>,
+        compiled_grammar: &mut Option<Box<dyn CompiledGrammar>>,
```
uuuvn, line 81: &Option is basically never a good idea (option doesn't own the grammar, it's an optional reference to a grammar, not a reference to an optional grammar). Should be`Option<&mut dyn CompiledGrammar>`

crates/uzu/Cargo.toml
```diff
@@ -39,15 +38,21 @@ objc2-foundation = { workspace = true, optional = true }
...
-criterion-macro.workspace = true
 rand_distr.workspace = true
 is_close.workspace = true
 serde_json5.workspace = true
 schemars.workspace = true
+
+[target.'cfg(not(target_arch = "wasm32"))'.dev-dependencies]
+proptest.workspace = true
+criterion.workspace = true
+criterion-macro.workspace = true
```
uuuvn: ?

crates/uzu/src/language_model/grammar/factory.rs
```diff
@@ -0,0 +1,43 @@
+#[cfg(grammar_xgram)]
```
uuuvn: This entire file should be done very differently. There should not be factories, just a function that returns Option<Box<dyn CompiledGrammar>>

uuuvn (changes requested):

crates/uzu/Cargo.toml
agolokoz: obviously this crates doesn't compile for wasm32

crates/uzu/src/language_model/grammar/factory.rs
agolokoz: why? why current approach is not good? what if `CompiledXGrammar::from_config` returns error?

crates/uzu/Cargo.toml
uuuvn: But it is in dev-dependencies and I do not see you also disable tests that use proptest so this does nothing? Just moves the failure from proptest doesn't compile to cannot find proptest when compiling tests. Tests working on wasi should be another pr

crates/uzu/src/language_model/grammar/factory.rs
uuuvn: Yes, it is very verbose for no good reason

.github/actions/setup/action.yml
```diff
@@ -18,30 +18,45 @@ inputs:
     description: "Install nextest"
     required: false
     default: 'false'
+  wasi:
+    description: "Install wasi"
+    required: false
+    default: 'false'
```
uuuvn, line 24: This should be a list of targets, not "install wasi"

.github/actions/setup/action.yml
```diff
@@ -18,30 +18,45 @@ inputs:
...
       run: |
         arch -arm64 brew install cmake
         xcodebuild -downloadComponent MetalToolchain
       shell: bash
+
+    - if: inputs.wasi == 'true'
+      run: |
+        arch -arm64 rustup target add wasm32-wasip1-threads --toolchain stable
+      shell: bash
+
```
uuuvn, line 51: This should be done via targets option in the ` dtolnay/rust-toolchain@stable` above

https://github.com/dtolnay/rust-toolchain?tab=readme-ov-file#inputs

(much cleaner, also will likely be faster, also afaik some rust nightly builds might not have some targets)

.github/workflows/tests.yml
```diff
@@ -70,31 +71,39 @@ jobs:
...
           cargo +stable nextest run --profile=ci --no-default-features \
             ${{ format('--features {0}', matrix.backend) || '' }} \
             --features tracing
 
+      - name: Build uzu for target wasi
+        if: ${{ matrix.target == 'wasm32-wasip1-threads' && matrix.backend == 'cpu' && !matrix.tracing }}
+        run: |
+          cargo +stable build --package uzu --no-default-features \
+            --target wasm32-wasip1-threads \
+
```
uuuvn, line 102: This should not be a separate step. Unify with "Build crates" by adding --target matrix.target when not default and --package uzu when wasi.

.github/workflows/tests.yml
```diff
@@ -58,6 +58,7 @@ jobs:
     strategy:
       matrix:
         backend: [ 'cpu', 'metal' ]
+        target: [ 'default', 'wasm32-wasip1-threads' ]
```
uuuvn, line 61: Image

Job name should have a target in it if not default.

Also currently a bunch of jobs like "metal + wasi + no tracing" spawn and do literally nothing. I think it makes sense to switch to include rather than a full cartesian product https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/run-job-variations#expanding-or-adding-matrix-configurations

Like we probably want to have:
- metal (build all crates, build benches for uzu)
- metal tracing (build all crates, run tests)
- cpu (build all crates)
- cpu + wasi (build uzu)

This can be done elegantly with include for specifying all matrix variants + some intelligently placed ifs and conditional arguments in run

crates/uzu/build/main.rs
```diff
@@ -25,14 +25,21 @@ async fn main() -> anyhow::Result<()> {
         println!("cargo::rerun-if-changed=/var/empty/hack_nonexistent_file_to_always_rerun");
     }
 
-    let metal_backend = cfg!(feature = "metal")
-        && matches!(env::var("CARGO_CFG_TARGET_OS").unwrap().as_ref(), "macos" | "ios" | "tvos" | "visionos");
+    let target_os = env::var("CARGO_CFG_TARGET_OS")?;
+    let target_arch = env::var("CARGO_CFG_TARGET_ARCH")?;
```
uuuvn, line 29: Usually arch first then os

crates/uzu/build/main.rs
```diff
@@ -25,14 +25,21 @@ async fn main() -> anyhow::Result<()> {
...
 
+    let metal_backend = cfg!(feature = "metal") && matches!(target_os.as_ref(), "macos" | "ios" | "tvos" | "visionos");
     println!("cargo::rustc-check-cfg=cfg(metal_backend)");
     if metal_backend {
         println!("cargo::rustc-cfg=metal_backend");
     }
 
+    let grammar_xgrammar = target_arch != "wasm32";
```
uuuvn, line 37: Let's have (default enabled) grammar feature so that grammar-disabled builds can be tested outside of wasm. Similar to how metal default-enabled feature exists rather than gating purely on arch/os

crates/uzu/src/language_model/grammar.rs
```diff
@@ -1,184 +1,42 @@
...
+pub fn create_compiled_grammar(
+    #[allow(unused)] config: Option<GrammarConfig>,
+    #[allow(unused)] tokenizer: &tokenizers::Tokenizer,
+    #[allow(unused)] stop_token_ids: Option<&[i32]>,
+) -> Result<Option<Box<dyn CompiledGrammar>>, Error> {
+    #[cfg(grammar_xgrammar)]
+    {
+        let Some(ref grammar_config) = config else {
+            return Ok(None);
         };
```
uuuvn, line 28: What's the point of this? Why would `create_compiled_grammar` even take an `Option` config with None never creating a grammar? Callers should do map instead.

crates/uzu/src/language_model/grammar.rs
```diff
@@ -1,184 +1,42 @@
...
-                trigger_token_id,
-                trigger_distance: None,
-            }
-        } else {
-            CompiledGrammarEngagementState::Always
+pub fn create_compiled_grammar(
+    #[allow(unused)] config: Option<GrammarConfig>,
+    #[allow(unused)] tokenizer: &tokenizers::Tokenizer,
+    #[allow(unused)] stop_token_ids: Option<&[i32]>,
+) -> Result<Option<Box<dyn CompiledGrammar>>, Error> {
```
uuuvn, line 23: Similarly to the previous comment, this should not return an `Option` but just error when the caller requests to have grammar in a build that doesn't have any grammar backends (now only xgrammar) enabled

crates/uzu/src/language_model/mod.rs
```diff
@@ -1,3 +1,5 @@
+mod grammar_xgrammar;
```
uuuvn, line 1: #[cfg(grammar_xgrammar)]

Also the module should just be named xgrammar. Cfg is namded grammar_xgrammar to avoid collision with a feature name to avoid confusion

crates/uzu/build/main.rs
uuuvn: Also I think it makes sense to do optional grammar in a separate prerequisite pr.

crates/uzu/src/data_type.rs
```diff
@@ -50,6 +52,7 @@ impl DataType {
     }
 }
 
+#[cfg(grammar_xgrammar)]
 impl From<DataType> for DLDataType {
     fn from(data_type: DataType) -> Self {
         Self {
```
uuuvn, line 58: This should probably be an extension-trait (like crates/uzu/src/backends/metal/metal_extensions/data_type.rs) in xgrammar file.

Cargo.toml
```diff
@@ -76,6 +76,7 @@ test-tag = "0.1"
 proptest = "1.10"
 criterion = "0.8.2"
 criterion-macro = "0.4.0"
+rustix = "1"
```
uuuvn, line 79: Why add a dependency?

uuuvn (changes requested):

Cargo.toml
uuuvn: Nightly requirement for wasm is fine, but non-wasm targets should build on stable

PR #335, opened 2026-04-10
Xgrammar optional

agolokoz:


.github/workflows/tests.yml
```diff
@@ -51,13 +51,14 @@ jobs:
...
   tests:
-    name: ${{ matrix.backend }} ${{ matrix.tracing && ' (tracing)' || '' }}
+    name: ${{ matrix.backend }} ${{ matrix.tracing && ' (tracing)' || '' }} ${{ matrix.grammar && ' (grammar)' || '' }}
     runs-on: [ self-hosted, macos, mac, arm64 ]
     if: ${{ github.event_name != 'pull_request' || github.event.pull_request.draft == false }}
 
     strategy:
       matrix:
         backend: [ 'cpu', 'metal' ]
+        grammar: [ false, true ]
```
uuuvn: We probably don't want to double the number of ci jobs just for grammar. Let's only add one job with cpu, no tracing and grammar: https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/run-job-variations#expanding-or-adding-matrix-configurations

.github/workflows/tests.yml
```diff
@@ -75,26 +76,29 @@ jobs:
...
             ${{ matrix.tracing && '--features tracing' || '' }} \
 
       - name: Run tests
         if: ${{ matrix.backend != 'cpu' && matrix.tracing }}
         run: |
           cargo +stable nextest run --profile=ci --no-default-features \
             ${{ format('--features {0}', matrix.backend) || '' }} \
-            --features tracing
+            --features grammar \
+            --features tracing \
```
uuuvn: This will run the same thing (grammar always enabled) twice for matrix.grammar = true and false

crates/uzu/src/language_model/grammar/xgram.rs
```diff
@@ -1,184 +1,41 @@
-use std::iter::repeat_n;
```
uuuvn, line 1: this should be grammar/mod.rs and xgrammar.rs should be grammar/xgrammar.rs

crates/uzu/src/session/chat_session.rs
```diff
@@ -259,8 +256,12 @@ impl ChatSession {
 
         let sampling_method = config.sampling_policy.resolve(language_model_config);
 
-        let mut compiled_grammar: Option<CompiledGrammar> = if let Some(ref grammar_config) = config.grammar_config {
-            Some(CompiledGrammar::from_config(grammar_config, None, &self.tokenizer_info)?)
+        let mut compiled_grammar: Option<Box<dyn CompiledGrammar>> = if let Some(ref config) = config.grammar_config {
+            match create_compiled_grammar(config, &self.tokenizer, Some(&self.stop_token_ids)) {
+                Ok(grammar) => Some(grammar),
+                Err(Error::GrammarNoBackend) => None,
+                Err(err) => return Err(err),
```
uuuvn: Trying to use grammar when no grammar backend is compiled should result in an error, not silent lack of grammar constraining.

uuuvn (changes requested):

crates/uzu/src/session/chat_session.rs
agolokoz: Task was to make xgrammar standalone feature and now you ask to fail when this feature is disabled. Then why is all this necessary? What about wasm builds?

crates/uzu/src/language_model/grammar/xgram.rs
agolokoz: `grammar/xgrammar.rs` will bring import conflicts inside `grammar` module

crates/uzu/src/session/chat_session.rs
uuuvn: I do not ask you to fail when this feature is disabled. I ask you to fail when the feature is disabled **and something tries to use it**. When uzu is compiled without the grammar support trying to use the grammar should return an error, not silently run without grammar while pretending that the decoding is grammar-constrained.

agolokoz: What is _trying to use the grammar_?

uuuvn: > What is `trying to use the grammar`?

crates/uzu/tests/integration/session/config/grammar_json_schema_test.rs

uuuvn:
```rust
fn test_grammar(speculator_config: SpeculatorConfig) {
    let model_dir = get_test_model_path();

    if !model_dir.exists() {
        eprintln!(
            "Skipping test: model not found at {}.\nRun scripts/download_test_model.sh first.",
            model_dir.display()
        );
        return;
    }

    let decoding_config = DecodingConfig::default()
        .with_sampling_seed(uzu::prelude::SamplingSeed::Custom(42))
        .with_speculator_config(speculator_config);
    let mut session = Session::new(model_dir, decoding_config).expect("Failed to create session");

    let input = Input::Text("Generate a detailed person profile with address and hobbies in JSON format.".to_string());

    let grammar_config = GrammarConfig::from_json_schema_type::<Person>().expect("Failed to create grammar config");

    let run_config =
        RunConfig::default().tokens_limit(1024).sampling_policy(SamplingPolicy::Default).grammar_config(grammar_config);

    let output =
        session.run(input, run_config, None::<fn(uzu::session::types::Output) -> bool>).expect("Failed to run session");

    let stats = output.stats;
    let total_tokens = stats.total_stats.tokens_count_input + stats.total_stats.tokens_count_output;

    println!("\n=== Test Results ===");
    println!("Generated output length: {} chars", output.text.original.len());
    println!(
        "Input tokens: {}, Output tokens: {}",
        stats.total_stats.tokens_count_input, stats.total_stats.tokens_count_output
    );
    println!("Total time: {:.2}s", stats.total_stats.duration);
    println!("Throughput: {:.2} tokens/s", total_tokens as f64 / stats.total_stats.duration);

    match serde_json::from_str::<Person>(&output.text.original) {
        Ok(person) => {
            println!("✓ Valid Person JSON!");
            println!("  Name: {}", person.name);
            println!("  Age: {}", person.age);
            println!("  Email: {}", person.email);
            println!("  Address: {}, {}", person.address.city, person.address.country);
            println!("  Hobbies: {:?}", person.hobbies);

            assert!(!person.name.is_empty());
            assert!(person.age > 0);
            assert!(person.email.contains('@'));
            assert!(!person.address.city.is_empty());
            assert!(!person.hobbies.is_empty());
        },
        Err(e) => {
            panic!("✗ Generated invalid JSON: {}\nOutput: {}", e, output.text.original);
        },
    }
}
```

Here is an example of grammar usage. The important lines are:

```rust
    let grammar_config = GrammarConfig::from_json_schema_type::<Person>().expect("Failed to create grammar config");
```

We create grammar config that constrains output of the llm to be a json that conforms to the Person schema

```rust
    let run_config = RunConfig::default().tokens_limit(1024).sampling_policy(SamplingPolicy::Default).grammar_config(grammar_config);
```

And in run config we do `.grammar_config(grammar_config)` attaching this to a RunConfig

uuuvn: After that the user of uzu api expects the output generated by an llm to be a valid json that conforms to the Person schema. This should fail loudly when uzu is compiled without grammar support rather than generating something that is likely wrong

uuuvn (approved):

PR #340, opened 2026-04-11
Faster quantized matmul for small suffix_len

CC-Yeh:
**Why it improves:** QmvFast reloads weights from memory once per output row — at small M (batch_dim) the kernel is bandwidth-bound and M identical weight reads dominate runtime. The BM=8 tile loads each weight
   tile into threadgroup memory **once** and reuses it across 8 rows via a simdgroup matmul, shifting the kernel from **bandwidth-bound to compute-bound**.

**Why BM=8 specifically:** Metal's simdgroup matmul works on 8×8 blocks, so BM=8 is the smallest tile that fills one block exactly. Bigger tiles pad the unused rows — e.g., BM=32 running on M=8 wastes 75% of
  the threads.

  ## Config

  | | |
  |---|---|
  | Hardware | Apple M4 Pro |
  | Dtype | BF16, 4-bit MLX, gs=128 |
  | N | {2048, 4096, 14336} |
  | K | {2048, 4096, 8192, 14336} |
  | Criterion | sample-size=10, warm-up=1s, measurement=2s |

  ## Speedup vs main dispatcher

  | M | avg | range | routing change |
  |---|---|---|---|
  | 6 | 1.01× | 0.97–1.04× | unchanged (QmvFast) |
  | **8** | **1.67×** | 1.29–2.01× | QmvFast → BM=8 |
  | **12** | **1.47×** | 1.31–1.68× | QmvFast → BM=8 |
  | **16** | **2.02×** | 1.79–2.26× | QmvFast → BM=8 |
  | **20** | **1.58×** | 1.36–1.79× | QmvFast → BM=8 |
  | **24** | **2.11×** | 1.85–2.34× | QmvFast → BM=8 |
  | **28** | **1.80×** | 1.59–2.03× | QmvFast → BM=8 |
  | **32** | **1.37×** | 1.19–1.56× | big tile → BM=8 |
  | **40** | **1.13×** | 0.97–1.32× | big tile → BM=8 |
  | 48 | 1.00× | 0.99–1.01× | unchanged (big tile) |
  | 56 | 1.00× | 0.99–1.01× | unchanged (big tile) |
  | 64 | 1.00× | 0.99–1.01× | unchanged (big tile) |

  Biggest wins at wide-N shapes (N=14336, K≥4096), smallest wins at narrow N=2048.


CC-Yeh:
In next PRs:

`VARIANTS_TUPLE` of sth similar to simplify the tile configs.

And probably can start discussing Autotuner, this is only tune on M4 and not exhausting the search space of possible tiles.

uuuvn (approved):

PR #341, opened 2026-04-12
Wasi build

agolokoz:


Cargo.toml
```diff
@@ -76,6 +76,7 @@ test-tag = "0.1"
 proptest = "1.10"
 criterion = "0.8.2"
 criterion-macro = "0.4.0"
+rustix = "1.1.4"
```
agolokoz, line 79: This crate needed for unified reading from file (`file_read_exact_at`). Moreover this crate already added in project as sub crate.

crates/uzu/Cargo.toml
```diff
@@ -39,17 +39,24 @@ objc2-foundation = { workspace = true, optional = true }
...
 serde_json5.workspace = true
 schemars.workspace = true
+
+# these crates pull in terminal/process/threading deps that don't support wasm
+[target.'cfg(not(target_arch = "wasm32"))'.dev-dependencies]
 comfy-table.workspace = true
+criterion.workspace = true
+criterion-macro.workspace = true
 indicatif.workspace = true
+proptest.workspace = true
```
agolokoz: `cargo build` returns errors without moving this dependencies under config

.github/workflows/tests.yml
```diff
@@ -74,28 +87,21 @@ jobs:
...
+        run: cargo +stable build --package uzu --no-default-features $FEATURE_FLAGS
+
       - name: Build crates
-        run: |
-          cargo +stable build --no-default-features \
-            ${{ matrix.backend != 'cpu' && format('--features {0}', matrix.backend) || '' }} \
-            ${{ matrix.grammar && '--features grammar' || '' }} \
-            ${{ matrix.tracing && '--features tracing' || '' }} \
+        if: ${{ matrix.target != 'wasm32-wasip1-threads' }}
+        run: cargo +stable build --no-default-features $FEATURE_FLAGS
```
uuuvn: These are almost identical except for the `--package uzu` part. Can you unify them by doing something like `${{ matrix.target == 'wasm32-wasip1-threads' && '--package uzu' || '' }}`?

crates/uzu/build/main.rs
uuuvn, line 38: `let grammar_xgrammar = cfg!(feature = "grammar") && target_arch != "wasm32";`

crates/uzu/Cargo.toml
uuuvn: This is very strange, dev-dependencies should not matter for -p uzu build. But if this is really true that it fails then sure, not worth investigating much for now

uuuvn (changes requested):

.github/workflows/tests.yml
```diff
@@ -50,21 +50,34 @@ jobs:
...
+          - backend: 'cpu'
+            tracing: true
 
     env:
       CARGO_ENCODED_RUSTFLAGS: "-Dwarnings"
+      FEATURE_FLAGS: >-
+        ${{ matrix.target != '' && format('--target {0}', matrix.target) || '' }}
+        ${{ matrix.backend != 'cpu' && format('--features {0}', matrix.backend) || '' }}
+        ${{ matrix.grammar && '--features grammar' || '' }}
+        ${{ matrix.tracing && '--features tracing' || '' }}
```
uuuvn, line 80: This is a very good cleanup btw

crates/uzu/Cargo.toml
agolokoz: You are right, with `-p uzu` it compiles successfully

crates/uzu/Cargo.toml
uuuvn, line 54: Whitespace change (just moves proptest, criterion, criterion-macro) (and into a less logical ordering)

uuuvn (changes requested):

uuuvn (approved):
Everything good except one small thing, pre-approved.

PR #343, opened 2026-04-14
wasi tests

agolokoz:


Cargo.toml
```diff
@@ -67,14 +67,14 @@ inquire = "0.9"
...
+comfy-table = { version = "7.2", default-features = false }
 autocxx = "0.30"
 cxx = "1.0"
 cmake = "0.1"
 autocxx-build = "0.30"
 test-tag = "0.1"
-proptest = "1.10"
-criterion = "0.8.2"
+proptest = { version = "1.10", default-features = false, features = ["std"] }
+criterion = { version = "0.8.2", default-features = false }
```
uuuvn, line 77: We probably want some proptest/criterion features other than std on non-wasi builds. Can you just enable them for non-wasm targets in uzu/Cargo.toml? Should be as simple as:

```toml
[target.'cfg(not(target_arch = "wasm32"))'.dev-dependencies]
criterion = { workspace = true, default-features = true }
proptest = { workspace = true, default-features = true }
```

(right after the dev-dependencies key please)

But in case you can't override default-features to true (I know you can't to false) you can try `features = ["default"]`

uuuvn (changes requested):

uuuvn (approved):

PR #360, opened 2026-04-21
Improve RHT + LoRA decode

CC-Yeh:
3.3% faster decode on LFM2.5-1.2B-RHT-QLoRA
  
  - Offline math trick: A_down' = A_down · H at load time
  - Fused A_down' into RMSNorm kernel (decode)
  - Fused A_up SG0-tail into QmvFast kernel (decode)
  - Per-rank dispatch (fused at r=16, unfused fallback otherwise)
  - One-line prefill recovery
  - LORA_RANK plumbed as kernel VARIANT
  - CPU LoRA reference + cross-backend tests
  - Shared adapter_up buffer (−22 MB)


crates/backend-uzu/src/backends/common/kernel/quant_matmul.rs
```diff
@@ -45,6 +46,9 @@ pub struct QuantizedMatmulArguments<'a, B: Backend> {
     pub output_buffer: &'a mut B::Buffer,
     pub hadamard_factors: Option<&'a B::Buffer>,
     pub batch_dim: usize,
+    pub h_buffer: Option<&'a B::Buffer>,
+    pub adapter_up: Option<&'a B::Buffer>,
+    pub lora_scale: f32,
```
uuuvn, line 51: Are these always used together/not used together? If so it should be another struct and one option of that structure. As much invariants (like `h_buffer.is_some() == adapter_up.is_some()`, `lora_scale` is not needed at not used when lora not enabled, etc) as possible should be expressed via type system (like via having one top option and inner struct).

crates/backend-uzu/src/encodable_block/linear/mod.rs
```diff
@@ -166,7 +180,7 @@ impl<B: Backend> dyn Linear<B> {
         parameter_tree: &ParameterTree<B::Context>,
         input_array_id: ArrayId,
         output_array_id: ArrayId,
-    ) -> Result<(Box<dyn Linear<B>>, Option<B::Buffer>), LinearBlockError<B>> {
+    ) -> Result<(Box<dyn Linear<B>>, Option<B::Buffer>, Option<LoraFusion<B>>), LinearBlockError<B>> {
```
uuuvn, line 183: Big scary tuple, let's make extracted fusions a struct. And struct can have nice helpers for extracting fusions and erroring if something was silently dropped. We can also do the same with declaring fusion capabilities (a struct of bools of what extracted fusions we can handle)

crates/backend-uzu/src/encodable_block/linear/qlora_wrapper.rs
```diff
@@ -46,34 +63,13 @@ pub struct QLoRALinearWrapper<B: Backend> {
...
     output_dim: usize,
     lora_rank: usize,
     lora_scale: f32,
     input_array_id: ArrayId,
     output_array_id: ArrayId,
-}
-
-// TODO: figure out how to make this generic over QLoRAWrapperError::InvalidTensor or make one global "Invalid Tensor" error and make this a common helper
-fn validate_tensor<'file, 'context, 'leaf, B: Backend>(
-    weights_leaf: &ParameterLeaf<'file, 'context, 'leaf, B::Context>,
```
uuuvn, line 60: The move ate the todo

uuuvn (changes requested):

ry2009:
Noticed a potential issue with the sign/Hadamard ordering in `compose_rotated_adapter_down`.

The precomposition needs `A_down' @ x == A_down @ H_rht(x)`, which requires applying `H_rht^T` to the rows of A_down, not `H_rht`.

Since `H_rht = H @ diag(s)` (signs first, then butterfly -- matching `simdgroup_random_hadamard_transform`), the transpose is `H_rht^T = diag(s) @ H` (butterfly first, then signs).

But `compose_rotated_adapter_down` calls `hadamard_kernel.encode()` which applies the standard `H @ diag(s)` (signs first)... this gives the wrong result.

Quick repro (JAX, same math):
- Apply `H @ diag(s)` to rows -> max error vs ground truth: **112.9**
- Apply `diag(s) @ H` to rows -> max error vs ground truth: **0.033** (f32 noise)

Fix: apply Hadamard butterfly to A_down rows first, then multiply by signs -- instead of the current order... lmk if this is intended though

CC-Yeh:
> Noticed a potential issue with the sign/Hadamard ordering in `compose_rotated_adapter_down`.
> 
> The precomposition needs `A_down' @ x == A_down @ H_rht(x)`, which requires applying `H_rht^T` to the rows of A_down, not `H_rht`.
> 
> Since `H_rht = H @ diag(s)` (signs first, then butterfly -- matching `simdgroup_random_hadamard_transform`), the transpose is `H_rht^T = diag(s) @ H` (butterfly first, then signs).
> 
> But `compose_rotated_adapter_down` calls `hadamard_kernel.encode()` which applies the standard `H @ diag(s)` (signs first)... this gives the wrong result.
> 
> Quick repro (JAX, same math):
> 
> * Apply `H @ diag(s)` to rows -> max error vs ground truth: **112.9**
> * Apply `diag(s) @ H` to rows -> max error vs ground truth: **0.033** (f32 noise)
> 
> Fix: apply Hadamard butterfly to A_down rows first, then multiply by signs -- instead of the current order... lmk if this is intended though

Not intended at all, thanks for catching that!

uuuvn:
@CC-Yeh what's the status of this pr? We definitely want it after the review fixes

CC-Yeh:
> @CC-Yeh what's the status of this pr? We definitely want it after the review fixes

Forgot to compute A_down for those layers can't be fused (not RMSNorm), the performance is the same as main after patching that, still trying to figure out a way to speed this up. Will spend 1-2 days more on this

CC-Yeh:
**TL;DR.** Runtime QLoRA costs ~10% decode / ~4% prefill on M4. Two fusion attempts gave 0% net. **Load-time merging the LoRA delta into the 4-bit base weights and re-quantizing fully recovers the tax** (verified n=15 × 2 benches), and is rank-independent. Recommend lalamo integrate calibrated merge for single-task deployments (if accuracy make sense ofc).

  ## Numbers (LFM2.5-1.2B, M4 base, n=15, 854→512)

  | Model | Decode t/s | Prefill t/s |
  |---|---|---|
  | RHT-4bit (no QLoRA, baseline) | 142.3 ± 1.6 | 1553 ± 3 |
  | RHT + QLoRA-r=32 (runtime) | 129.0 ± 2.2 | 1490 ± 6 |
  | **RHT + QLoRA-r=32 merged** | **144.0 ± 1.6** | **1555 ± 2** |
  | **RHT + QLoRA-r=64 merged** | **144.1 ± 1.3** | **1554 ± 3** |
  | **RHT + QLoRA-r=128 merged** | **144.3 ± 1.6** | **1554 ± 3** |

  Tax: **−9.4% decode / −4.1% prefill**. Merge: tied with baseline at every rank.

  (Caveat: our local r=64/r=128 are zero-padded from r=32, so the merged W_eff is bit-identical across ranks. Theory says a real high-rank
  fine-tune would also tie since the merged kernel has zero rank-dependent state.)

  ## Where the 9.4% comes from

  Per decode token, base RHT-4bit reads 670 MB at 95 GB/s realized → 7.05 ms ≈ 142 t/s. Adding LoRA at r=32 adds **40 MB of adapter weights**
  and **128 small dispatches per token** (64 A_down + 64 A_up GEMVs). Measured: 7.75 ms (= 129 t/s); the 40 MB transferred at only ~57 GB/s —
  about 60% of the base kernel's effective bandwidth. **That bandwidth degradation is the entire QLoRA tax.**

  ## Why small adapter kernels run at lower bandwidth

  DRAM has high latency (~300 ns on M-class chips). To hit peak bandwidth you need many memory requests in flight simultaneously. The number
  of in-flight requests scales with how many threadgroups are concurrently resident on each shader core.

  - **Base kernel** (`mlp_up`, 19 MB read): hundreds of TGs → memory channel near peak (~95 GB/s)
  - **Adapter kernel** (`A_down` at r=32, 256 KB read): ~32 TGs total → channel sits half-idle (~57 GB/s)

  It's a parallelism problem, not a transfer-rate problem. Small kernels can't issue enough concurrent demand to keep DRAM saturated. The only
   way out is to make the kernel bigger — which is what fusion attempts.

  ## What we tried, and why it didn't help

  **RMSNorm + A_down kernel fusion** — *predicted +1-2%, measured 0%.*
  Folding `h_partial[r=32]` into RMSNorm as per-thread state increases register usage. On Apple Silicon, register usage determines occupancy
  in discrete jumps — crossing a threshold halves concurrent threadgroups per shader core. The fused kernel's collapsed occupancy cancels the
  launch overhead saved. Zero-sum.

  **Full kernel fusion (A_down + A_up into base)** — *should help, measured −10%.*
  Each of the 64 base threadgroups would need the full `h`, but `h` is computed by a different threadgroup. Without grid-wide synchronization
  (Apple doesn't expose CUDA's cooperative-groups equivalent), the only option is to recompute A_down 64 times. Compute dominates the dispatch
   savings.

  Same pattern both times: making the runtime kernel bigger to amortize launch overhead either trips the occupancy cliff or requires a Metal
  primitive that doesn't exist.

  ## How merge works

  **The technique:** precompute `W_eff = W_q + s · B · A` once at load time, re-quantize back to 4-bit, drop the adapter tensors. Math-wise

  That's why every cost goes to zero: 0 adapter dispatches (vs 128), 0 adapter bytes per token (vs 40 MB), 0 rank-axis state in any kernel.
  Rank is also irrelevant post-merge — the merged kernel has nothing that depends on r.

  Verified the engine actually executes the merged weights (generation diverges from baseline at token 4, matches the QLoRA fine-tune's
  stylistic signature). Bonus: naive RTN re-quant produced coherent English, so calibrated re-quant will likely preserve fine-tune quality
  well.

PR #363, opened 2026-04-23
Kv cache sparse buffers

agolokoz:


crates/backend-uzu/src/backends/common/encoder.rs
```diff
@@ -133,6 +133,10 @@ impl<'encoding, B: Backend> Encoder<'encoding, B> {
             allocation_pool: self.allocation_pool,
         }
     }
+
+    pub fn end_submit_wait(self) -> Result<Completed<B>, B::Error> {
+        self.end_encoding().submit().wait_until_completed()
+    }
```
uuuvn, line 139: This should not be a thing (trivial, "end encoding, submit, and wait" is a very strange function to have)

crates/backend-uzu/src/utils/sparse_array.rs
```diff
@@ -0,0 +1,323 @@
...
+    #[error("Failed to create Encoder: {0}")]
+    CreateEncoderError(B::Error),
+}
+
+#[derive(Debug)]
+pub struct SparseArray<B: Backend> {
+    buffer: Rc<RefCell<B::SparseBuffer>>,
+    data_type: DataType,
+    shape: Box<[usize]>,
+}
```
uuuvn, line 25: This should not exist. It's either Array<Backend, Buffer> (so which buffer is being wrapped is also a generic) or use raw B::SparseBuffer, or something else, but definitely not this.

In general it is **never** a good idea to copy paste code and if you ever find yourself in a situation where you either copy paste one thing and change it a bit or write something that looks very close to a piece of code that already exist you should stop and find a way to unify them

crates/backend-uzu/tests/unit/backends/common/kernel/kv_cache_update_test.rs
```diff
@@ -115,32 +97,45 @@ fn test_random_pattern(context: &<Metal as Backend>::Context) {
...
 }
 
+#[test]
+fn test_kv_cache_update_kernel() {
+    let metal_context = match <Metal as Backend>::Context::new() {
+        Ok(ctx) => ctx,
+        Err(e) => {
+            println!("Failed to create MetalContext: {:?}. Skipping test.", e);
+            return;
+        },
```
uuuvn: I see you didn't write it and this is just a move of an existing test, but please fix things like that when you modify them. Tests should never be silently skipped

uuuvn (changes requested):
Skimmed through, left a couple comments on things.

One global question is whether the layout change is good perf wise or not. Can you measure perf practically and look theoretically (ask ai, it's good enough at this stuff) into what the memory access patters of heavy kernels accessing those buffers (not all of them matter equally) look like?

If not we'll have to make sparse buffer api more complicated to accommodate existing layout, but it's better than losing perf

There also should be a non-sparse path for backends that don't support it (webgpu, cpu should be opt-in for debugging or not at all maybe, idk)

crates/backend-uzu/src/backends/common/encoder.rs
agolokoz: It's just a syntax sugar. I am tired to write `end_encoding().submit().wait_until_completed()` every time.

You also write "In general it is never a good idea to copy paste code"

crates/backend-uzu/src/utils/sparse_array.rs
agolokoz: There is no copy paste. `SparseArray` is not `Array`, they have different behavior. And it's not possible to use raw B::SparseBuffer. Suggest anything else.

crates/backend-uzu/src/backends/common/sparse_buffer.rs
```diff
@@ -0,0 +1,20 @@
+use std::fmt::Debug;
+
+use crate::backends::common::Backend;
+
+pub trait SparseBuffer: Debug {
```
CC-Yeh, line 5: If sparse array exists only for Metal, maybe we don't want it be implemented for every backend?

crates/backend-uzu/src/backends/metal/sparse_buffer.rs
```diff
@@ -0,0 +1,137 @@
...
+    capacity: usize,
+    length: usize,
+}
+
+impl MetalSparseBuffer {
+    pub fn new(
+        context: &MetalContext,
+        capacity: usize,
+    ) -> Result<MetalSparseBuffer, MetalError> {
+        let page_size = MTLSparsePageSize::KB16;
```
CC-Yeh, line 40: I'm curious why `KB16` not `64` or `256`?

smaller page size does save some tail waste but also need to extend more frequently.

crates/backend-uzu/src/backends/metal/sparse_buffer.rs
```diff
@@ -0,0 +1,137 @@
...
+        };
+
+        let heap_desc = MTLHeapDescriptor::new();
+        heap_desc.set_type(MTLHeapType::Placement);
+        heap_desc.set_storage_mode(MTLStorageMode::Private);
+        heap_desc.set_size(aligned_capacity);
+        heap_desc.set_max_compatible_placement_sparse_page_size(page_size);
+        let Some(heap) = context.device.new_heap_with_descriptor(&heap_desc) else {
+            return Err(MetalError::SparseHeapAlloc(aligned_capacity, page_size_bytes));
+        };
```
CC-Yeh, line 59: Can you explain or provide some links on how SparseHeap work from Metal 4?

- Does this allocate full size heap eagerly? or lazily?

crates/backend-uzu/src/backends/metal/sparse_buffer.rs
CC-Yeh: Also should this be configurable?

crates/backend-uzu/src/backends/metal/sparse_buffer.rs
CC-Yeh: If this allocate full size heap eagerly, can we really save memory usage?

crates/backend-uzu/src/backends/metal/sparse_buffer.rs
```diff
@@ -0,0 +1,137 @@
...
+
+    fn buffer_mut(&mut self) -> &mut <Self::Backend as Backend>::Buffer {
+        &mut self.buffer
+    }
+
+    fn capacity(&self) -> usize {
+        self.capacity
+    }
+
+    fn extend(
```
CC-Yeh, line 104: Do we need boundary check if it would exceed capacity limit?

crates/backend-uzu/src/backends/metal/sparse_buffer.rs
```diff
@@ -0,0 +1,137 @@
...
+    ) {
+        let value = self.sync_counter.fetch_add(1, Ordering::Relaxed) + 1;
+        self.queue.update_buffer_mappings_heap_operations_count(
+            &self.buffer,
+            Some(&self.heap),
+            NonNull::new(operations.as_ptr() as *mut _).unwrap(),
+            operations.len(),
+        );
+        self.queue.signal_event_value(self.sync_event.as_ref(), value);
+        self.sync_event.wait_until_signaled_value_timeout_ms(value, u64::MAX);
```
CC-Yeh, line 85: finite timeout with error propogation?

crates/backend-uzu/src/backends/metal/sparse_buffer.rs
CC-Yeh: Would `extend_pages` help the call sites and this function be more maintainable? I see many transitions between bytes and pages.

crates/backend-uzu/src/backends/metal/sparse_buffer.rs
```diff
@@ -0,0 +1,137 @@
...
+            queue: context.command_queue4.clone(),
+            sync_event: context.device.new_shared_event().unwrap(),
+            sync_counter: AtomicU64::new(0),
+            page_size,
+            capacity: aligned_capacity,
+            length: 0,
+        })
+    }
+
+    fn execute(
```
CC-Yeh, line 73: This is called 2 times (K, V) buffers per layer in a forward pass.

My understanding is while GPU is running for current token, the CPU can encode next one in parallel.

But the `wait_until_signaled_value_timeout_ms` here block CPU and wait until GPU finish generating current token and update the buffer(my assumption please double check). Will this create pipeline bubble every time it needs a new page? If this is true, can we wait on GPU queue instead?

CC-Yeh (commented):
Do we want to support Metal 3? If so, we will need fallback.

Also regarding access pattern, I think write to buffer will be better in general but reading for GQA could be worse, can you check benchmarks for single-pass and two-pass attention kernels? especially GQA config.

  old:  KV cache shape = [num_groups, seq_len, head_dim]          (group-major)
  new:  KV cache shape = [seq_len, num_groups, head_dim]          (token-major)

crates/backend-uzu/src/backends/metal/sparse_buffer.rs
CC-Yeh: Ah I got it now, it is not saving the amount of memory allocated but more about reducing memory fragmentation etc.

Still curious about how Metal allocate and manage the heap though.

crates/backend-uzu/src/backends/metal/sparse_buffer.rs
uuuvn: The hw is also likely to group 16kb pages inside those bigger "software" pages together for less translation overhead, we definitely want to use a big page size

crates/backend-uzu/src/backends/common/sparse_buffer.rs
uuuvn: Yeah, we don't want it to be on every backend, I wrote about this in my review comment above

crates/backend-uzu/src/backends/metal/sparse_buffer.rs
uuuvn: It's not about fragmentation, it's about not consuming all the memory for the entire model context if you're going to ask it what's 2 + 2 or something trivial like that. We want kv cache memory consumption to grow with context, not preallocate full. This is super important for phones and for very long context models

crates/backend-uzu/src/backends/common/encoder.rs
uuuvn: Duplicating logic is bad, there is no logic, it's just replacing (). with _ and shortening end_encoding to end + wait_until_completed to just wait. Like this is not a coherent thing that you factor into a function, this is replacing .do_a().do_b().do_c() with a newly added .do_a_do_b_do_c().

crates/backend-uzu/src/utils/sparse_array.rs
uuuvn: Ughm, just open both side by side. there is so much copy paste. Like 30-40% can be trivially deleted under Array<Backend, Buffer> with SparseBuffer as is, and a bit more can (and should, I think this is a much better approach) be unified if Buffer is a supertrait of SparseBuffer with extra functions on SparseBuffers instead of a separate entity giving you an underlying Buffer like it's done in apple's metal apis.

It's also obviously possible to use raw sparse buffer there, that's just probably not a good idea with all the verbosity that would bring and the unified buffers thing I describe above as a superior option

crates/backend-uzu/src/backends/metal/sparse_buffer.rs
CC-Yeh: @uuuvn thanks for explanation, then I'm wondering how efficient does Metal serve new pages on request, would be interesting to see how it perform. (Hope they have a pool of free pages or something similar to reduce latency)

PR #366, opened 2026-04-24
Reduce Matmul LOC

eugenebokhan:


eugenebokhan:
This impl showed a perf degradation, closing the PR

PR #379, opened 2026-04-29
Sparse buffers

agolokoz:
- Adds a `SparseBuffer` trait to the common backend abstraction with `map`/`unmap` operations over page ranges, plus a `create_sparse_buffer` method on Context.
- Implements the `Metal` backend: `MetalSparseBuffer` backed by `MetalSparseHeaps` managed by a per-context `MetalSparseHeapPool` (256KB pages), with support for mapping multiple buffers into a single heap, and an `MTL4CommandQueue` used for sparse residency updates.
- Adds a CPU backend stub for `SparseBuffer`.
- Includes unit tests for `MetalSparseBuffer` and `MetalSparseHeapPool` (mapping, unmapping, freeing, multi-buffer packing, error paths).


eugenebokhan (approved):

crates/backend-uzu/src/backends/common/buffer.rs
```diff
@@ -1,13 +1,15 @@
...
+
 use crate::backends::common::Backend;
 
 pub trait Buffer: Debug {
     type Backend: Backend;
 
     fn gpu_ptr(&self) -> usize;
 
-    fn size(&self) -> usize;
+    fn size(&self) -> ByteSize;
```
uuuvn: @agolokoz @eugenebokhan if you want this it must be a separate pr, it's very hard to review already complicated prs that contain unrelated changes

crates/backend-uzu/src/backends/metal/error.rs
```diff
@@ -22,4 +25,12 @@ pub enum MetalError {
...
     #[error("Cannot create pipeline state: {0}")]
     CannotCreatePipelineState(String),
+    #[error("Can not allocate buffer with size={0}")]
+    SparseBufferAlloc(usize),
+    #[error("Can not allocate heap with size={0} and page size={1}")]
+    SparseHeapAlloc(usize, ByteSize),
+    #[error("Queue for sparse buffer is not available")]
+    SparseQueueNotAvailable,
+    #[error("Sparse buffer requires Metal 4 queue")]
+    SparseRequireMtl4Queue,
```
uuuvn: There is no such thing as "sparse queue", also why is "SparseRequireMtl4Queue" not "CannotCreateCommandQueueMtl4"?

crates/backend-uzu/src/backends/metal/sparse/mod.rs
```diff
@@ -0,0 +1,3 @@
+mod sparse_buffer;
+
+pub use sparse_buffer::MetalSparseBuffer;
```
uuuvn: What is the purpose of this file?

crates/backend-uzu/src/backends/metal/metal_extensions/sparse_resource_mapping_mode_extensions.rs
```diff
@@ -0,0 +1,21 @@
...
+}
+
+impl From<MTLSparseTextureMappingMode> for SparseResourceMappingMode {
+    fn from(mode: MTLSparseTextureMappingMode) -> Self {
+        match mode {
+            MTLSparseTextureMappingMode::Map => SparseResourceMappingMode::Map,
+            MTLSparseTextureMappingMode::Unmap => SparseResourceMappingMode::Unmap,
+        }
+    }
+}
```
uuuvn: Where is this used?

crates/backend-uzu/src/backends/metal/metal_extensions/library_extensions_pipeline.rs
```diff
@@ -1,5 +1,4 @@
-use metal::{MTLComputePipelineState, MTLDeviceExt, MTLFunctionConstantValues, MTLLibrary, MTLLibraryExt};
-use objc2::{rc::Retained, runtime::ProtocolObject};
+use metal::prelude::*;
```
uuuvn: Let's keep explicit imports, I think importing * is a bad practice because it makes it hard to see where something with a generic name (like Retained) actually came from

crates/backend-uzu/src/array/sparse.rs
```diff
@@ -0,0 +1,6 @@
+use crate::{
+    Array,
+    backends::common::{Backend, SparseBuffer},
+};
+
+impl<B: Backend, Buf: SparseBuffer<Backend = B>> Array<B, Buf> {}
```
uuuvn: What's the point of this code?

uuuvn (changes requested):

crates/backend-uzu/src/backends/metal/metal_extensions/sparse_resource_mapping_mode_extensions.rs
eugenebokhan: Unmap is not used anywhere for now but this is definitely better then have if else statements

crates/backend-uzu/src/backends/metal/metal_extensions/library_extensions_pipeline.rs
eugenebokhan: Can you give an example of an ambiguous situation of not knowing where a metal / objc type did come from?

crates/backend-uzu/src/backends/metal/sparse/mod.rs
eugenebokhan: This is an umbrella file gating what types actually go public and managing internal declarations.

crates/backend-uzu/src/backends/metal/error.rs
eugenebokhan: I would suggest a generic `SparseBufferCreationFailed`

crates/backend-uzu/src/backends/common/buffer.rs
eugenebokhan: It was quite handy to add it here as Metal api already operates sometimes in bytes, sometimes in kilobytes

crates/backend-uzu/src/array/sparse.rs
agolokoz: To have Array on SparseBuffer

crates/backend-uzu/src/backends/metal/sparse/mod.rs
uuuvn: This is what `pub` on structs/functions/etc is for, what's the point of having a something/something.rs that has a bunch of (noop) pubs and then an extra something/mod.rs that just reexports one actually-pub things and negates other pubs? This is either a useless indirection or a Rube Goldberg machine of reexports instead of just not marking things we don't want to be public as pub

crates/backend-uzu/src/backends/metal/metal_extensions/sparse_resource_mapping_mode_extensions.rs
uuuvn: image

I mean the MTLSparseTextureMappingMode => SparseResourceMappingMode path. When do we need to convert the mtl thing to generic thing? I don't think we do

crates/backend-uzu/src/backends/metal/metal_extensions/library_extensions_pipeline.rs
uuuvn: Retained or literally anything else that doesn't start with "MTL". If you don't already know it's very confusing, I think * are a bad practice in general for that reason (and not just in rust, any good linter will scream at you for doing `from something import *` in python too), it makes tracing where something came from harder. Though this is relatively benign, if everybody else likes `use something::prelude::*` I'm kind of okay with it, though I don't think we did that before in this codebase

crates/backend-uzu/src/array/sparse.rs
uuuvn: It's empty, it's not doing anything

crates/backend-uzu/src/backends/common/buffer.rs
uuuvn: I only see this:

image


Which isn't quite "metal api already operates sometimes in kilobytes". In this particular case ByteSize is nice but I'm skeptical that using it will make the code *overall* more readable because there are few cases where it makes things better and a lot of otherwise just noise .as_u64() everywhere. Doing the bytesize in a separate pr will make it much clearer whether we want it or not, here it's mixed in with a lot of other changes and hard to judge

crates/backend-uzu/src/array/sparse.rs
uuuvn: You're not implementing a marker trait or something like that here. It's an inherent impl with zero things inside of it, it's not doing anything at all and it's the only thing in this file. If there is no sparse-specific functionality yet it shouldn't have it's own useless/placeholder file

crates/backend-uzu/src/backends/common/sparse/sparse_buffer.rs
```diff
@@ -0,0 +1,59 @@
...
+
+use super::{SparseBufferOperation, SparseResourceMappingMode};
+use crate::backends::common::{Backend, Buffer};
+
+pub trait SparseBuffer: Buffer<Backend: Backend<SparseBuffer = Self>> {
+    fn get_mapped_pages(&self) -> &RangeMap<usize, ()>;
+
+    fn get_page_size(&self) -> ByteSize;
+
+    /// `operations.range` must be in pages
```
CC-Yeh: Inconsistent with comment in `mtl4_update_sparse_buffer_mapping_operation_extensions.rs`

crates/backend-uzu/src/backends/metal/context.rs
```diff
@@ -60,11 +56,13 @@ impl Context for MetalContext {
 
     fn new() -> Result<Rc<Self>, MetalError> {
         let device: Retained<ProtocolObject<dyn MTLDevice>> =
-            <dyn metal::MTLDevice>::system_default().ok_or(MetalError::CannotOpenDevice)?;
+            <dyn MTLDevice>::system_default().ok_or(MetalError::CannotOpenDevice)?;
 
         let command_queue =
             device.new_command_queue_with_max_command_buffer_count(1024).ok_or(MetalError::CannotCreateCommandQueue)?;
 
+        let command_queue4 = device.new_mtl4_command_queue().ok_or(MetalError::CannotCreateCommandQueueMtl4)?;
```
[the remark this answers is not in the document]
CC-Yeh: @agolokoz will `new()` work with Metal 3?

crates/backend-uzu/src/backends/common/sparse/sparse_buffer.rs
```diff
@@ -0,0 +1,59 @@
...
+        &mut self,
+        context: &<Self::Backend as Backend>::Context,
+        bytes_ranges: &[Range<usize>],
+    ) -> Result<(), <Self::Backend as Backend>::Error> {
+        let page_size_bytes = self.get_page_size().as_u64() as usize;
+        let pages_ranges = bytes_ranges
+            .iter()
+            .map(|bytes_range| Range {
+                start: bytes_range.start / page_size_bytes,
+                end: bytes_range.end / page_size_bytes,
```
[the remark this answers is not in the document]
CC-Yeh: Should we use `div_ceil` for `end` since the `end` is an exclusive boundary?

crates/backend-uzu/tests/common/helpers.rs
```diff
@@ -14,16 +14,7 @@ pub fn alloc_buffer_with_data<B: Backend, T: bytemuck::NoUninit>(
     context: &B::Context,
     data: &[T],
 ) -> B::DenseBuffer {
-    if data.len() == 0 {
-        // Metal doesn't allow creating 0-byte buffers, create a minimal buffer instead
-        return context.create_buffer(1).expect("Failed to create buffer");
-    }
```
CC-Yeh: Should we keep this `0` guard?

crates/backend-uzu/src/backends/metal/metal_extensions/mtl4_update_sparse_buffer_mapping_operation_extensions.rs
```diff
@@ -0,0 +1,10 @@
+use metal::prelude::*;
+
+use crate::backends::common::SparseBufferOperation;
+
+impl From<&SparseBufferOperation> for MTL4UpdateSparseBufferMappingOperation {
+    fn from(op: &SparseBufferOperation) -> Self {
+        // `op.range` is a byte range; the heap offset matches the buffer offset.
+        Self::new(op.mode.into(), op.range.clone(), op.range.start)
```
CC-Yeh: Is it okay to assume heap offset will always be the same with sparse buffer's offset?
assertion/guard could probably help and serve as documentation.

crates/backend-uzu/src/backends/metal/sparse/sparse_buffer.rs
```diff
@@ -0,0 +1,111 @@
...
+            return Err(MetalError::SparseBufferAlloc(aligned_capacity));
+        };
+
+        let heap_desc = MTLHeapDescriptor::new();
+        // Sparse buffers must be backed by a Placement heap with a sparse page size set;
+        // `MTLHeapType::Sparse` is for sparse textures and trips a runtime assertion when
+        // passed to updateBufferMappings.
+        heap_desc.set_type(MTLHeapType::Placement);
+        heap_desc.set_storage_mode(MTLStorageMode::Private);
+        heap_desc.set_size(aligned_capacity);
```
CC-Yeh: I'm curiously if this would allocate memory lazily and sparsely like `MTLHeapType::Sparse`?

CC-Yeh: If `capacity` is small and expected to grow backing heaps for sparsebuffer, how does it grow?

crates/backend-uzu/src/backends/metal/context.rs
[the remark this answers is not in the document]
CC-Yeh: Can we make `command_queue4` optional

CC-Yeh (changes requested):

uuuvn:
image

image


@CC-Yeh @agolokoz Can't reply to this for whatever github reason, no we should not add any metal 3 fallbacks, we already require metal 4 in crates/backend-uzu/build/metal/toolchain.rs (compile with min os version 26 and std metal 4)

crates/backend-uzu/src/backends/metal/context.rs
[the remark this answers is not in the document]
CC-Yeh: Per @uuuvn , we only want to support metal4, then is it necessary to have two command queues?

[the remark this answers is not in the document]
agolokoz: > @agolokoz will new() work with Metal 3?

No, but we support Metal at least 4 version

crates/backend-uzu/src/backends/metal/sparse/sparse_buffer.rs
agolokoz: Sparse is only for textures. Using Sparse here will bring crash in runtime

agolokoz: > If capacity is small and expected to grow backing heaps for sparsebuffer, how does it grow?

`MetalSparseHeapPool` added

crates/backend-uzu/src/backends/metal/context.rs
```diff
@@ -111,11 +136,37 @@ impl Context for MetalContext {
         &self,
         size: usize,
     ) -> Result<Retained<ProtocolObject<dyn MTLBuffer>>, MetalError> {
-        let buffer =
-            self.device.new_buffer(size, MTLResourceOptions::STORAGE_MODE_SHARED).ok_or(MetalError::CannotCreateBuffer);
+        // Metal doesn't allow creating 0-byte buffers, create a minimal buffer instead
+        let size = size.max(1);
```
uuuvn: Who creates 0-byte buffer and why? "0 byte buffer" is nonsensical, it should remain a hard error instead of having a workaround to allow a broken caller to do nonsensical things

crates/backend-uzu/src/backends/metal/context.rs
```diff
@@ -111,11 +136,37 @@ impl Context for MetalContext {
...
+
+    fn create_buffer_with_data(
+        &self,
+        data: &[u8],
+    ) -> Result<Retained<ProtocolObject<dyn MTLBuffer>>, MetalError> {
+        let data = if data.len() == 0 {
+            &[0]
+        } else {
+            data
+        };
```
uuuvn: Who creates 0-byte buffer and why? "0 byte buffer" is nonsensical, it should remain a hard error instead of having a workaround to allow a broken caller to do nonsensical things

crates/backend-uzu/src/backends/metal/sparse/sparse_buffer.rs
```diff
@@ -0,0 +1,115 @@
...
+        common::{Backend, Buffer, SparseBuffer},
+        metal::{Metal, error::MetalError, metal_extensions::SparsePageSizeExt},
+    },
+    prelude::MetalContext,
+};
+
+#[derive(Debug)]
+pub struct MetalSparseBuffer {
+    buffer: Retained<ProtocolObject<dyn MTLBuffer>>,
+    mapped_pages: RangeMap<usize, ()>,
```
uuuvn: Why is this a map with value () instead of a set?

crates/backend-uzu/src/backends/metal/sparse/sparse_buffer.rs
```diff
@@ -0,0 +1,115 @@
...
+    context: Weak<MetalContext>,
```
uuuvn: Why is this still weak?

crates/backend-uzu/src/backends/metal/sparse/sparse_buffer.rs
```diff
@@ -0,0 +1,115 @@
...
+}
+
+impl MetalSparseBuffer {
+    pub(crate) fn new(
+        context: Weak<MetalContext>,
```
uuuvn: Why is this still weak?

crates/backend-uzu/src/backends/metal/sparse/sparse_buffer.rs
```diff
@@ -0,0 +1,115 @@
...
+        capacity: usize,
+        page_size: MTLSparsePageSize,
+    ) -> Result<Self, MetalError> {
+        let ctx = context.upgrade().ok_or(MetalError::CannotCreateBuffer)?;
+        let page_size_bytes = page_size.in_bytes();
+        let aligned_capacity = capacity.next_multiple_of(page_size_bytes);
```
uuuvn: used once and very small, `capacity.next_multiple_of(page_size.in_bytes())` is much easier to read

crates/backend-uzu/src/backends/metal/sparse/sparse_heap.rs
```diff
@@ -0,0 +1,129 @@
...
+}
+
+impl MetalSparseHeap {
+    pub fn new(
+        context: &MetalContext,
+        capacity_bytes: usize,
+        page_size: MTLSparsePageSize,
+    ) -> Result<Self, MetalError> {
+        let page_size_bytes = page_size.in_bytes();
+        let aligned_capacity = capacity_bytes.next_multiple_of(page_size_bytes);
```
uuuvn: used once and very small, `capacity.next_multiple_of(page_size.in_bytes())` is much easier to read

crates/backend-uzu/src/backends/metal/sparse/sparse_buffer.rs
```diff
@@ -0,0 +1,115 @@
...
+        })
+    }
+}
+
+impl Drop for MetalSparseBuffer {
+    fn drop(&mut self) {
+        let context = self.context.upgrade().expect("Failed to upgrade context");
+        for (range, _) in self.mapped_pages.iter() {
+            context.sparse_heap_pool_mut().unmap(&context, &self.buffer, &range).expect("Failed to unmap");
+        }
```
CC-Yeh: I'm new to Rust. Is it a good idea to panic in destructor? iirc, this is not a good idea in c++.

crates/backend-uzu/src/backends/metal/sparse/sparse_heap_pool.rs
```diff
@@ -0,0 +1,165 @@
...
+                    heap.execute(buffer, &context.command_queue4, &[op_params], true);
+                    self.heaps.push(heap);
+                    pages_to_map.start += map_pages_count;
+                },
+                Err(err) => {
+                    // it's necessary to unmap previously mapped pages in existing heaps
+                    for (heap_pos, mappings) in existing_heaps_mappings {
+                        self.heaps[heap_pos].execute(buffer, &context.command_queue4, &mappings, false);
+                    }
+                    return Err(err);
```
CC-Yeh, line 100: Should we also drop mapped pages from line 91? those pages are not added to `existing_heaps_mapping`, possibly causing leak.

crates/backend-uzu/src/backends/metal/sparse/sparse_heap.rs
```diff
@@ -0,0 +1,129 @@
...
+            .new_heap_with_descriptor(&heap_desc)
+            .ok_or(MetalError::SparseHeapAlloc(aligned_capacity, page_size.in_bytes()))?;
+        Ok(Self {
+            heap,
+            buffer_mappings: HashMap::new(),
+        })
+    }
+
+    /// Heap page ranges within `heap_range` that have no buffer mapped to them.
+    pub fn free_pages_in(
```
CC-Yeh: How about keep a `free_pages` variable instead of reconstruct it every time.
for one `map()`, this costs `O(H x B x M x log(BxM))` just for free-page computation

H= heaps, B = buffers mapped to this heap, M = mappings per buffer

crates/backend-uzu/src/backends/metal/sparse/sparse_heap.rs
```diff
@@ -0,0 +1,129 @@
...
+            .into_iter()
+            .flat_map(|m| m.iter().map(|(r, mapping)| (r.clone(), mapping)))
+    }
+
+    pub fn is_empty(&self) -> bool {
+        self.buffer_mappings.is_empty()
+    }
+
+    /// Command queue calls doesn't have any synchronization.
+    /// It is the responsibility of the caller.
```
CC-Yeh, line 78: Is it possible to enforce syncrhonization by submitting this ops to encode queues?

crates/backend-uzu/src/backends/metal/sparse/sparse_heap.rs
```diff
@@ -0,0 +1,129 @@
...
+                let mode = if map {
+                    MTLSparseTextureMappingMode::Map
+                } else {
+                    MTLSparseTextureMappingMode::Unmap
+                };
+                MTL4UpdateSparseBufferMappingOperation::new(mode, op.buffer_pages.clone(), op.heap_page_offset)
+            })
+            .collect();
+
+        cmd_queue.update_buffer_mappings(buffer, Some(&self.heap), &mtl_operations);
```
CC-Yeh, line 102: How often will this get called for a typical kv cache usage? If it is too many calls, maybe we could batch it to reduce overhead.

crates/backend-uzu/tests/unit/backends/metal/sparse/sparse_buffer_test.rs
```diff
@@ -0,0 +1,203 @@
...
+fn create_context() -> Rc<<Metal as Backend>::Context> {
+    <Metal as Backend>::Context::new().expect("Failed to create Metal context")
+}
+
+/// Capacity sized to fit `heap_count` worth of heaps.
+fn buffer_capacity(
+    ctx: &<Metal as Backend>::Context,
+    heap_count: usize,
+) -> usize {
+    heap_count * ctx.sparse_heap_pool_mut().heap_capacity_bytes()
```
CC-Yeh: all `sparse_heap_pool_mut` usage in test looks like could be read-only, does not need to be mutable

uuuvn (changes requested):

CC-Yeh (commented):
Looks good 🚀 , left some comments to be discussed

Can we have a brief summary of changes in PR body?
Also some benchmarks would be great to see if this really works.

crates/backend-uzu/src/backends/metal/sparse/sparse_buffer.rs
uuuvn: No, but it's better than silently ignoring an error. eprintln doesn't count either, nobody will see it.

In the future we should have something like sentry / other global logger to just log this stuff

crates/backend-uzu/src/backends/metal/sparse/sparse_heap.rs
agolokoz: `SparseHeap::execute` (this function) already supports batching. Which batching do you mean?

crates/backend-uzu/src/backends/metal/sparse/sparse_heap.rs
agolokoz: Seems no. The only function to work with mappings is `MTL4CommandQueue::update_buffer_mappings`

crates/backend-uzu/src/backends/metal/sparse/sparse_heap.rs
CC-Yeh: across heaps, is it possible?

agolokoz:
> Also some benchmarks would be great to see if this really works.

Which benchmarks do you want to see? I can suggest benchmarks for memory using, but we know that memory used only for allocated heaps

CC-Yeh:
> > Also some benchmarks would be great to see if this really works.
> 
> Which benchmarks do you want to see? I can suggest benchmarks for memory using, but we know that memory used only for allocated heaps

Memory usage one is certainly good, I'm thinking about long context attention so we can try to see how growing KV cache would impact the performance.

agolokoz:
> Memory usage one is certainly good, I'm thinking about long context attention so we can try to see how growing KV cache would impact the performance.

Migration KV cache to sparse buffers and benchmarks will be in the future PR

CC-Yeh (approved):

crates/backend-uzu/src/backends/metal/sparse/sparse_heap.rs
agolokoz: Multiple mappings available only for pair 1 buffer - 1 heap. Maybe in future PR I will add a function that takes multiple buffers, heaps and operations and join operations for the same pairs.

crates/backend-uzu/src/backends/metal/sparse/sparse_heap.rs
```diff
@@ -0,0 +1,128 @@
...
+        free_pages.insert(0..(aligned_capacity / page_size.in_bytes()));
+        Ok(Self {
+            heap,
+            buffer_mappings: HashMap::new(),
+            free_pages,
+        })
+    }
+
+    /// Heap page ranges within `heap_range` that have no buffer mapped to them.
+    pub fn free_pages_in<'a>(
```
uuuvn: Should be just free_pages, the only caller uses this as "calculate heap size and pass as heap_range". This should not be the responsibility of a caller

crates/backend-uzu/src/backends/metal/sparse/sparse_heap.rs
```diff
@@ -0,0 +1,128 @@
...
+                };
+                MTL4UpdateSparseBufferMappingOperation::new(mode, op.buffer_pages.clone(), op.heap_page_offset)
+            })
+            .collect();
+
+        cmd_queue.update_buffer_mappings(buffer, Some(&self.heap), &mtl_operations);
+
+        let buffer_address = buffer.gpu_address();
+        let entry = self.buffer_mappings.entry(buffer_address).or_default();
+        let free_pages = &mut self.free_pages;
```
uuuvn: ???

crates/backend-uzu/src/backends/metal/sparse/sparse_heap_pool.rs
```diff
@@ -0,0 +1,166 @@
...
+    page_size: MTLSparsePageSize,
+}
+
+impl MetalSparseHeapPool {
+    pub fn new(
+        page_size: MTLSparsePageSize,
+        heap_capacity: usize,
+    ) -> Self {
+        Self {
+            heaps: vec![],
```
uuuvn: Vec::new()

crates/backend-uzu/src/backends/metal/context.rs
```diff
@@ -111,11 +143,37 @@ impl Context for MetalContext {
         &self,
         size: usize,
     ) -> Result<Retained<ProtocolObject<dyn MTLBuffer>>, MetalError> {
-        let buffer =
-            self.device.new_buffer(size, MTLResourceOptions::STORAGE_MODE_SHARED).ok_or(MetalError::CannotCreateBuffer);
+        if size == 0 {
+            return Err(MetalError::ZeroBuffer);
+        }
+
```
uuuvn: What does metal do if you try to allocate zero-sized buffer? Might be worth it if it panics, but otherwise MetalError::CannotCreateBuffer is sufficient

crates/backend-uzu/src/backends/metal/context.rs
```diff
@@ -111,11 +143,37 @@ impl Context for MetalContext {
...
+    }
+
+    fn create_buffer_with_data(
+        &self,
+        data: &[u8],
+    ) -> Result<Retained<ProtocolObject<dyn MTLBuffer>>, MetalError> {
+        if data.len() == 0 {
+            return Err(MetalError::ZeroBuffer);
+        }
+
```
uuuvn: What does metal do if you try to allocate zero-sized buffer? Might be worth it if it panics, but otherwise MetalError::CannotCreateBuffer is sufficient

crates/backend-uzu/Cargo.toml
```diff
@@ -94,7 +94,6 @@ metal = [
     "dep:clang-ast",
     "dep:objc2",
     "dep:objc2-foundation",
-    "dep:bytesize",
     "dep:metal",
 ]
```
uuuvn, line 99: ?

uuuvn (changes requested):

uuuvn (approved):

PR #382, opened 2026-05-01
Dense buffers

agolokoz:


crates/backend-uzu/src/backends/common/buffer.rs
```diff
@@ -1,34 +1,31 @@
-use std::{fmt::Debug, ops::Range, os::raw::c_void, ptr::NonNull};
+use std::{fmt::Debug, ops::Range};
 
-use super::Backend;
+use backend_uzu::backends::common::DenseBuffer;
 
 pub trait Buffer: Debug {
-    type Backend: Backend<Buffer = Self>;
```
uuuvn, line 6: Buffer is still fundamentally attached to a certain backend, do not remove `type Backend: Backend` from `trait Buffer`, just move the bound to DenseBuffer like `trait DenseBuffer: Buffer<Backend: Backend<DenseBuffer = Self>>` or something along those lines

crates/backend-uzu/src/backends/common/buffer.rs
```diff
@@ -1,34 +1,31 @@
...
     );
-
-    fn cpu_ptr(&self) -> NonNull<c_void>;
-    fn gpu_ptr(&self) -> usize;
-
-    fn length(&self) -> usize;
 }
 
-pub trait BufferGpuAddressRangeExt: Buffer {
+pub trait BufferGpuAddressRangeExt: DenseBuffer {
```
uuuvn: What's the meaning of this change exactly?

crates/backend-uzu/src/backends/common/buffer.rs
```diff
@@ -1,34 +1,31 @@
-use std::{fmt::Debug, ops::Range, os::raw::c_void, ptr::NonNull};
+use std::{fmt::Debug, ops::Range};
 
-use super::Backend;
+use backend_uzu::backends::common::DenseBuffer;
```
uuuvn: self

uuuvn (changes requested):

crates/backend-uzu/src/backends/common/buffer.rs
uuuvn: Can you gate src/lib.rs reexport hack behind a cfg to avoid this in the future? (separate pr)

```
// needed for tests to resolve `backend_uzu::` imports
extern crate self as backend_uzu;
```

uuuvn (approved):

PR #384, opened 2026-05-04
Sparse Buffers Update

eugenebokhan:


crates/backend-uzu/src/backends/common/sparse_resources/sparse_buffer.rs
```diff
@@ -54,29 +52,29 @@ pub trait SparseBufferExt: SparseBuffer {
...
     }
 }
 
+#[cfg(metal_backend)]
 #[derive(Debug)]
 pub(crate) struct SparseBufferMappedPages {
     map: RangeMap<usize, ()>,
 }
 
+#[cfg(metal_backend)]
```
agolokoz: Maybe it's better to keep only 1 such cfg in `mod.rs` for easier support?

crates/backend-uzu/src/backends/metal/sparse_buffer/sparse_buffer_mapped_pages.rs
```diff
@@ -0,0 +1,30 @@
+use rangemap::RangeMap;
+
+use crate::backends::common::{SparseBufferOperation, SparseResourceMappingMode};
+
+#[derive(Debug)]
+pub(super) struct SparseBufferMappedPages {
```
agolokoz: This struct (and SparseBuffer) will be used also for Vulkan, I think it should be putted back to common module

crates/backend-uzu/src/backends/common/sparse_resources/sparse_buffer_operation.rs
```diff
@@ -0,0 +1,22 @@
+use std::ops::Range;
+
+use super::SparseResourceMappingMode;
+
+/// A single sparse-buffer page-mapping operation. `range` is expressed in
+/// **bytes**, page-aligned to the owning buffer's `MTLSparsePageSize`.
+pub struct SparseBufferOperation {
```
agolokoz, line 7: Page aligned bytes range will bring more complexity to code that will use `SparseBuffer`: clients must do extra calculations, although `SparseBufferExt::map_bytes` implements this logic. Do we really want strict range with page aligned bytes?

eugenebokhan: I only introduced the mapping mode type for more clarity, the original operation type was [here](https://github.com/trymirai/uzu/blob/663c438fb94d0e05140d4c297c8276bb3f7b167d/crates/backend-uzu/src/backends/common/sparse_buffer.rs#L7)

agolokoz (approved):

PR #385, opened 2026-05-04
Codegen: Optional Expression Rewriter

eugenebokhan:
## What changed

- `OPTIONAL(...)` conditions like `OPTIONAL(quant_method == QuantizationMethod::AWQ)` now are available to use.
- Replaced `use_zero_points` / `use_mlx_quant` bool pair in `Qmv`, `QmvFast`, `QmmTransposed` kernels with single `quant_method: QuantizationMethod` enum.
- Split bindgen unreadable spaghetti into a typed sum-of-products module.

## Bindgen decoupling

```
build/metal/bindgen/
├── mod.rs                       # orchestrator
├── arguments.rs                 # ArgumentEmission { Buffer | Constant | IndirectDispatch }
├── specialize.rs                # SpecializeEmission
├── variants.rs                  # VariantBind
├── variant_path_rewriter.rs     # variant ident → self.field rewriter
├── dispatch.rs                  # DispatchEmission (axis / direct / indirect)
└── trait_wiring.rs              # TraitWiring
```

Each kernel argument is parsed once into a typed variant; each emission stream is one `iter().filter_map()` pass:

```rust
pub enum ArgumentEmission {
    Buffer(BufferArgument), 
    Constant(ConstantArgument), 
    IndirectDispatch(IndirectDispatchArgument),
}

impl ArgumentEmission {
    pub fn struct_field(&self)               -> Option<TokenStream>;
    pub fn struct_initializer(&self)         -> Option<TokenStream>;
    pub fn encode_argument_definition(&self) -> TokenStream;
    pub fn encode_lifetime(&self)            -> Option<TokenStream>;
    pub fn encode_deconstruct(&self)         -> Option<TokenStream>;
    pub fn encode_access(&self)              -> Option<TokenStream>;
    pub fn encode_set(&self)                 -> TokenStream;
}
```

Replaced a `.multiunzip()` over a 7-tuple of `Vec<Option<TokenStream>>` and three `mut` accumulators (`arg_count`, `indirect_flag`, `referenced_variants`) with typed fields and a `VariantPathRewriter` consumed via `.finish()`.

`mod.rs` orchestrator — `parse → emit → assemble`:

```rust
let variant_binds         = variants::parse(kernel)?;
let argument_emissions    = arguments::parse(kernel, enum_path_rewriter)?;
let specialize_emission   = specialize::parse(kernel, base_index, kernel_name)?;
let trait_wiring          = trait_wiring::build(kernel, &trait_name, &struct_name);
let mut variant_rewriter  = VariantPathRewriter::new(&variant_binds, kernel_name);
let dispatch_emission     = dispatch::parse(kernel, &mut variant_rewriter)?;
let referenced            = variant_rewriter.finish();

// One pass per output stream
let conditional_buffer_fields: Vec<_> =
    argument_emissions.iter().filter_map(|a| a.struct_field()).collect();
// …etc

quote! { /* assemble struct + impl from the named pieces */ }
```


crates/backend-uzu/build/metal/enum_path_rewrite.rs
```diff
@@ -0,0 +1,141 @@
...
+        &self,
+        condition: &str,
+    ) -> String {
+        let parsed = match syn::parse_str::<Expr>(condition) {
+            Ok(expr) => expr,
+            Err(_) => return format!("static_cast<bool>({condition})"),
+        };
+        match emit_metal_expr(&parsed, self) {
+            Ok(body) => format!("static_cast<bool>({body})"),
+            Err(_) => format!("static_cast<bool>({condition})"),
```
CC-Yeh: Should we let it fail earlier instead of getting cryptic metal compile error?

crates/backend-uzu/src/backends/metal/kernel/generated/quantization_method.h
```diff
@@ -0,0 +1,12 @@
+// Auto-generated from gpu_types/quantization_method.rs - do not edit manually
+#pragma once
+
+#include <metal_stdlib>
+using namespace metal;
+
+namespace uzu::quantization_method {
+enum class QuantizationMethod : uint32_t {
+  MLX = 0,
```
CC-Yeh: How about `ZeroPoint` and `Bias`?

crates/backend-uzu/build/metal/wrapper.rs
```diff
@@ -46,7 +75,7 @@ fn kernel_wrappers(
                 .filter(|a| matches!(a.argument_type(), Ok(MetalArgumentType::Specialize(_))))
                 .enumerate()
                 .map(|(i, a)| {
-                    let c_type = a.c_type.trim_start_matches("const ");
+                    let c_type = specialize_constant_type(&a.c_type);
```
CC-Yeh: Curious why we use #define/#undef here instead of having rewrite_for_metal substitute bare specialize names with their prefixed counterparts, wouldn't that keep all name resolution in one place?

crates/backend-uzu/src/backends/metal/metal_extensions/function_constant_values_extensions_set_value.rs
```diff
@@ -22,6 +24,10 @@ impl FunctionConstantValueType for u32 {
     const DATA_TYPE: MTLDataType = MTLDataType::UInt;
 }
 
+impl FunctionConstantValueType for QuantizationMethod {
+    const DATA_TYPE: MTLDataType = MTLDataType::UInt;
+}
+
```
uuuvn: This is not the right way to do it, every backend shouldn't implement it's own lowering of enums to uints

uuuvn (changes requested):

crates/backend-uzu/src/backends/metal/metal_extensions/function_constant_values_extensions_set_value.rs
uuuvn: The point of having a huge build system with dsl is to have stuff like this be handled by it, not by manually writings this. I think a good way to do this is just convert all the enums to u32 before attaching in bindgen so that metal never sees the enum type, just a u32

uuuvn (approved):

PR #386, opened 2026-05-04
Array buffer split

agolokoz:
1. Split array.rs into multiple files
2. Add 2nd generic parameter `BUF: Buffer<Backend = B>` to Array with default `<B as Backend>::DenseBuffer`


crates/backend-uzu/src/array/common.rs
```diff
@@ -0,0 +1,85 @@
+use std::{cell::RefCell, rc::Rc};
+
+use crate::{
+    DataType,
+    array::size_for_shape,
+    backends::common::{Backend, Buffer},
+};
+
+#[derive(Debug)]
+pub struct Array<B: Backend, BUF: Buffer<Backend = B> = <B as Backend>::DenseBuffer> {
```
eugenebokhan:
```suggestion
pub struct Array<B: Backend, Buffer: Buffer<Backend = B> = <B as Backend>::DenseBuffer> {
```

crates/backend-uzu/src/array/common.rs
```diff
@@ -0,0 +1,85 @@
...
+    pub(super) buffer: Rc<RefCell<BUF>>,
+    pub(super) offset: usize,
+    pub(super) shape: Box<[usize]>,
+    pub(super) data_type: DataType,
+}
+
+impl<B: Backend, BUF: Buffer<Backend = B>> Array<B, BUF> {
+    // Constructors
+    pub unsafe fn from_parts(
+        buffer: Rc<RefCell<BUF>>,
```
eugenebokhan: Can we consider lifetime ref here instead of Rc since this is a subview of the original buffer?

crates/backend-uzu/src/array/common.rs
```diff
@@ -0,0 +1,85 @@
...
+        self.offset
+    }
+
+    pub fn shape(&self) -> &[usize] {
+        &self.shape
+    }
+
+    pub fn data_type(&self) -> DataType {
+        self.data_type
+    }
```
eugenebokhan: Why have funcs since the properties are public? If we a satisfying the trait constraints, then better make the props private

crates/backend-uzu/src/array/common.rs
```diff
@@ -0,0 +1,85 @@
...
+
+    pub fn size(&self) -> usize {
+        size_for_shape(&self.shape, self.data_type)
```
eugenebokhan: Maybe better to store it as a pre-calculated value given the non-mutable nature of this type

crates/backend-uzu/src/array/util.rs
```diff
@@ -0,0 +1,17 @@
+use crate::DataType;
+
+pub fn size_for_shape(
+    shape: &[usize],
+    data_type: DataType,
+) -> usize {
+    let Some(last_dim) = shape.last() else {
+        return data_type.size_in_bytes();
```
eugenebokhan, line 8: Can we consider usage of the ByteSize type here as well as the return type?

crates/backend-uzu/src/array/common.rs
agolokoz: To prevent intersections with `trait Buffer` I used `BUF` generic. Changed to `Buf`

crates/backend-uzu/src/array/common.rs
agolokoz: But Array for SpraseBuffer probably will have mutable size (depends on how we decide to do it ofc)

crates/backend-uzu/src/array/common.rs
agolokoz: pub(super) makes fields public only for parent module. 

I restored Array and all impls to one file, to keep fields private.

crates/backend-uzu/src/array/common.rs
agolokoz: Can not. A lot of code is tied to `fn buffer(&self) -> Rc<RefCell<BUF>>`

crates/backend-uzu/src/array/util.rs
agolokoz: Too much code changes will be involved by this change. I prefer to made this change in separate PR

eugenebokhan (approved):

uuuvn (approved):
Approved with 3 changes:

- [simplified `ArrayContextExt`](https://github.com/trymirai/uzu/pull/386/changes/3630ee4617f029dd07ab93a60588461c15cee4ed) (preexisting thing, you just moved the code that was already bad and I noticed in the diff, that bad part is almost definitely written by me in the past)
- [renamed what you've added as context.rs to have a more accurate context_ext.rs name](https://github.com/trymirai/uzu/pull/386/changes/bbba7c6837233c0e97187c7d962b5ff0e014cffb) (this is not the actual context but an extension and the file name should ideally reflect that)
- [moved dense array impl into a separate file](https://github.com/trymirai/uzu/pull/386/changes/fc46643d3d53589b1edb2317948a2610bcdc846a) (your earlier version of this pr I've briefly looked at had this, not sure why you reverted this change, it was good)

PR #390, opened 2026-05-06
Support per-layer RoPE config format

Madfyre:
Summary
- Add support for the new per-layer RoPE config format used by converted models
- Preserve compatibility with existing single RoPE config models
- Add coverage for generated RoPE buffer construction


crates/backend-uzu/src/forward_pass/state/rope_buffers.rs
```diff
@@ -15,12 +19,14 @@ pub struct RopeBuffers<B: Backend> {
 impl<B: Backend> RopeBuffers<B> {
     pub fn new(
         context: &B::Context,
+        rope_config: &RoPEConfig,
         model_shape: &ModelShape,
     ) -> Self {
-        let rope_dim = model_shape.rope_dim();
-        let rope_max_sequence_length = model_shape.context_length();
+        let common = rope_config.common();
+        let rope_dim = common.head_dim.unwrap_or_else(|| model_shape.rope_dim());
```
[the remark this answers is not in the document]
Madfyre: @codex review

crates/backend-uzu/src/config/decoder.rs
```diff
@@ -141,6 +141,13 @@ impl<'de> Deserialize<'de> for DecoderConfig {
             None
         };
         let layer_types_value = explicit_layer_types.or(derived_layer_types);
+        let (global_rope_config, local_rope_config) = resolve_rope_configs(
+            global_rope_config,
+            local_rope_config,
+            &layer_config_value,
+            layer_configs_boxed.as_deref(),
+        )
+        .map_err(de::Error::custom)?;
```
uuuvn: Why are you trying to guess what is local and what is global with heuristics and why is there a cpu fallback fill? There is no longer local/global distinction, there are just different ropes and they can be whatever, multiple global, multiple local, etc.

uuuvn (changes requested):

crates/backend-uzu/src/config/decoder.rs
Madfyre: My main goal was to keep minimalistic PR

So, I saw that in `shared_buffers.rs` we have `global_rope` and `local_rope` and all backend modeled as these two shared buffers! So when the new lalamo format added rope_config on each layer, I treated it as a compatibility problem: derive the old `global_rope_config / local_rope_config` fields from the new per-layer configs. That is why I wrote `resolve_rope_configs`! This made like a small adapter that would let the rest of the existing Uzu runtime keep working.

But I agree with you! The model should represent distinct RoPE configs directly and map layers to those configs, without guessing global/local.

I also checked the current lalamo exporter and local converted artifacts. New per-layer RoPE exports do not appear to serialize RoPE tensors: `Transformer.export_weights()` only exports `layers` and `output_norm`, while RoPE tables are initialized from `RoPEConfig` using `rc.init()`. Also, the local Gemma-4 safetensors header also does not contain rope tensors. That's why I made config-based RoPE initialization.

Madfyre: I’ll push a new commit with your suggested approach this evening.

crates/backend-uzu/src/forward_pass/state/rope_type.rs
```diff
@@ -2,6 +2,7 @@ use serde::{Deserialize, Serialize};
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize, Deserialize)]
 pub enum RopeType {
+    Layer(usize),
     Global,
     Local,
 }
```
uuuvn: There is no longer global/local rope

uuuvn (changes requested):

Madfyre:
What changed:
- Removed shared RoPE cosine/sine buffers. RoPE is now computed directly from each layer's `RoPEConfig` inside the CPU/Metal kernels instead of materializing `[max_sequence_length, rope_dim]` tables.
- Per-layer `rope_config` is passed directly into `Rope`; legacy `global_rope_config` / `local_rope_config` remains only as a compatibility fallback for older configs.
- Removed the `RopeType::{Global, Local, Layer}` runtime model.

One question:
I do not have enough experience here, so I wanted to double-check the direction. Gemma 4 RoPE seems a bit special because partial RoPE uses the full head dimension for frequency exponents / pair layout. Would it make sense to encode this explicitly in lalamo, similar to how Llama has its own RoPE config, instead of making Uzu infer that behavior from `partial_rotary_dim < head_dim`?

Madfyre:
Since the previous commit, this adds a small RoPE hot-path optimization. The measured performance impact is close to neutral, but it keeps the RoPE kernel simpler by moving inverse-frequency computation out of the hot path.

RoPE still computes `sin/cos` on the fly, so we avoid materializing large `[max_sequence_length, rope_dim]` tables. The change is that `inverse_frequencies` are now precomputed once per `Rope` block into a tiny buffer and passed into the CPU/Metal kernels.

This removes the repeated `base/scaling/log/exp` frequency calculation from the inner RoPE kernel loop.

crates/backend-uzu/src/backends/cpu/kernel/rope/rope.rs
```diff
@@ -7,47 +7,78 @@ use crate::ArrayElement;
...
     rope_dim: u32,
+    rotary_pair_stride: u32,
+    inverse_frequency_count: u32,
+    rope_max_sequence_length: u32,
+    rope_attention_scaling_factor: f32,
     num_heads: u32,
     num_groups: u32,
     suffix_length: u32,
-    max_sequence_length: u32,
 ) {
```
uuuvn, line 62: Why is this needed?

uuuvn (changes requested):

crates/backend-uzu/src/backends/cpu/kernel/rope/rope.rs
Madfyre: `inverse_frequencies` replaces the old `cosines`/`sines` tables: the kernel computes `angle = position * inverse_frequency`, then derives cos/sin inside the kernel.

So this changes the kernel input from precomputed position-by-dimension tables to config-derived frequencies. That is why the buffer is much smaller: it stores only `rope_dim / 2` frequencies instead of full `[max_sequence_length, rope_dim]` tables.

I thought precomputing the cos/sin tables would be more reasonable than recomputing them every time inside the RoPE kernel!

`rotary_pair_stride` is needed because this implementation stores partial RoPE compactly as `rope_dim = partial_rotary_dim`, while lalamo applies partial RoPE in the full `head_dim` rotate-half layout. 

For example, with `head_dim = 512` and `partial_rotary_dim = 128`, lalamo masks the inverse frequencies so only the first 64 frequency slots are active, then duplicates them across the two halves of the full head. That means the affected dimensions are `0..63` and `256..319`, not a contiguous `0..127` prefix!

`inverse_frequency_count` is only a safety guard for the raw pointer kernel interface, so the kernel can avoid reading past the provided frequency buffer.

`rope_max_sequence_length` is the RoPE config sequence bound used by the existing position handling.

`rope_attention_scaling_factor` is needed for scaled RoPE variants

uuuvn: 1). This is completely unrelated to the new config format and if this was a good change it should've been a separate pr
2). It's actually not a good change, small differences in precision of sin/cos between reference implementations and metal/cpu/whatever cause output quality degradation which is why we do what we do. https://www.lesswrong.com/w/chesterton-s-fence

PR #392, opened 2026-05-07
Sparse heaps

agolokoz:


agolokoz:
@codex review

crates/backend-uzu/src/backends/metal/context.rs
```diff
@@ -69,6 +84,10 @@ impl Context for MetalContext {
 
         let device_capabilities = MetalDeviceCapabilities::from_device(&device);
 
+        let page_size = MTLSparsePageSize::KB256;
+        let heap_capacity = 64 * 4 * page_size.byte_size().as_u64() as usize;
```
CC-Yeh, line 88: Why 64MB? Any trade-offs?

crates/backend-uzu/src/backends/metal/sparse/sparse_heap.rs
```diff
@@ -0,0 +1,86 @@
...
+                let mode = if map {
+                    MTLSparseTextureMappingMode::Map
+                } else {
+                    MTLSparseTextureMappingMode::Unmap
+                };
+                MTL4UpdateSparseBufferMappingOperation::new(mode, op.buffer_pages.clone(), op.heap_page_offset)
+            })
+            .collect();
+
+        cmd_queue.update_buffer_mappings(buffer, Some(&self.heap), &mtl_operations);
```
CC-Yeh, line 78: potential race? Do we need fence/barrier for intra/inter queue cases?

crates/backend-uzu/src/backends/metal/context.rs
CC-Yeh: Maybe we could make this configurable and smaller for iphones, bigger for macs

crates/backend-uzu/src/backends/metal/sparse/sparse_utils.rs
```diff
@@ -0,0 +1,43 @@
+use std::ops::Range;
+
+#[derive(Clone, PartialEq)]
+pub(super) struct MetalSparseHeapBufferMapping {
+    gpu_address: u64,
```
CC-Yeh, line 5: Could `gpu_address` be brittle if `Drop` fails or finished partially?
Maybe store a pointer to buffer? Or we could store the mapping on buffer side?

crates/backend-uzu/src/backends/metal/sparse/sparse_heap.rs
```diff
@@ -0,0 +1,86 @@
...
+    pub fn mapped_pages(&self) -> &RangeMap<usize, MetalSparseHeapBufferMapping> {
+        &self.mapped_pages
+    }
+
+    pub fn execute(
+        &mut self,
+        buffer: &ProtocolObject<dyn MTLBuffer>,
+        cmd_queue: &ProtocolObject<dyn MTL4CommandQueue>,
+        operations: &[MetalSparseHeapMappingParameters],
+        map: bool,
```
CC-Yeh, line 60: How about using enum to be more explicit.

crates/backend-uzu/src/backends/metal/sparse/sparse_buffer.rs
```diff
@@ -15,56 +20,49 @@ use crate::{
...
+    fn drop(&mut self) {
+        let Some(context) = self.context.upgrade() else {
+            return;
+        };
 
-        Ok(())
+        let ranges: Vec<Range<usize>> = self.mapped_pages.iter().map(|(range, _)| range.clone()).collect();
+        ranges.iter().for_each(|range| {
+            let _ = context.sparse_heap_pool_mut().unmap(&context, &self.buffer, range);
+        });
```
CC-Yeh, line 66: should we not dropping errors?

crates/backend-uzu/src/backends/metal/sparse/sparse_heap_pool.rs
```diff
@@ -0,0 +1,156 @@
...
+                };
+                mappings.push(mapping);
+
+                pages_to_map.start += map_pages_count;
+                if pages_to_map.len() == 0 {
+                    break;
+                }
+            }
+
+            heap.execute(buffer, &context.command_queue4, &mappings, true);
```
CC-Yeh, line 70: early return when no overlapping to skip expensive op?

CC-Yeh (commented):
What do you think about having a warm pool of heaps to reduce TTFT? we can trade some memory for latency. Might be useful for short inputs.

agolokoz:
> What do you think about having a warm pool of heaps to reduce TTFT? we can trade some memory for latency. Might be useful for short inputs.

Planned to do in the future

crates/backend-uzu/src/backends/metal/context.rs
agolokoz: > Maybe we could make this configurable and smaller for iphones, bigger for Macs

Or even for different cases we can have different pools

Sizes will be set up together with first usage of sparse buffers. For now I just add some value

crates/backend-uzu/src/backends/metal/sparse/sparse_heap.rs
agolokoz: Caller must be responsible for synchronizations because in case of many maps, for example, it's not good idea to synchronize every map

crates/backend-uzu/src/backends/metal/sparse/sparse_utils.rs
agolokoz: > Could gpu_address be brittle if Drop fails or finished partially?

How it's possible?
Anyway after drop() call object is not usable anymore.

Or I didn't understand what you mean.

crates/backend-uzu/src/backends/metal/sparse/sparse_heap.rs
agolokoz: Enum will have only 2 values with possibility to have more or less, but we don't need it. For me enum is more strict.

crates/backend-uzu/src/backends/metal/sparse/sparse_heap_pool.rs
agolokoz: There are nothing expensive inside if mappings are empty

crates/backend-uzu/src/backends/metal/sparse/sparse_heap.rs
CC-Yeh: Thanks, how can we enforce this? Maybe tests + documentation can help

crates/backend-uzu/src/backends/metal/sparse/sparse_heap_pool.rs
CC-Yeh: My understanding is that cpu will talk to GPU even if mappings is empty, maybe this is cheap I'm not sure.

crates/backend-uzu/src/backends/metal/sparse/sparse_utils.rs
CC-Yeh: The pool keys is bookkeeping by `gpu_address`, a value Metal can reassign to a new buffer later, so any orphan entry left behind by a partial Drop becomes indistinguishable from a live entry.

But in case of duplicate entries, probably first unmap would release the heap and second one is a no-op.

CC-Yeh (approved):
LGTM overall, let's also get a pass from @uuuvn

crates/backend-uzu/src/backends/metal/sparse/sparse_utils.rs
agolokoz: For your case I have only 1 idea: right after new sparse buffer creation, remove all pages associated with new buffer `gpu_address`. How do you think is it enough?

crates/backend-uzu/src/backends/metal/sparse/sparse_heap_pool.rs
agolokoz: Will add checking for empty list

PR #401, opened 2026-05-12
Buffer ranges refactoring

agolokoz:
* `BufferRangeRef` and `BufferRangeMut` are moved to separate file
* Removed redundant lifetimes
* Subrange calculation moved to separate function to prevent copypaste


uuuvn (approved):
Good pr, thanks!

PR #403, opened 2026-05-13
Allocations clean up

LuckyIYI:
Moved variables closer to where they are used
Separated scratch from inputs allocations, removing them from arguments


crates/backend-uzu/src/backends/common/kernel/moe/experts_two_pass_decode.rs
```diff
@@ -49,63 +49,57 @@ impl<B: Backend> MoeExpertsTwoPassDecodeBlock<B> {
...
         let h_blocks = (args.d_ff as u32 + BLOCK_M - 1) / BLOCK_M;
-        self.pass_a_tile.encode_counts(
+        let tile_counts = self.pass_a_tile.encode_counts(
             encoder,
             MoePassATileCountsArguments {
                 expert_offsets: args.expert_offsets,
-                tile_counts: &mut *args.tile_counts,
                 e: args.e,
                 h_blocks,
             },
```
uuuvn: Very small struct, I think it's going to be cleaner if the 3 arguments are just arguments of the function instead of one argument with a 3-field struct

crates/backend-uzu/src/backends/common/kernel/moe/experts_two_pass_decode.rs
```diff
@@ -49,63 +49,57 @@ impl<B: Backend> MoeExpertsTwoPassDecodeBlock<B> {
...
+        )?;
+
+        let tile_scan = self.pass_a_tile.encode_scan(
             encoder,
             MoePassATileScanArguments {
-                tile_counts: &*args.tile_counts,
-                tile_offsets: &mut *args.tile_offsets,
-                total_tiles: &mut *args.total_tiles,
+                tile_counts: &tile_counts,
                 e: args.e,
```
uuuvn: Very small struct, I think it's going to be cleaner if the 2 arguments are just arguments of the function instead of one argument with a 2-field struct

Same in a bunch of places below too, won't spam comments on each one

uuuvn (changes requested):

uuuvn (approved):

PR #404, opened 2026-05-13
M5 macOS 26.5 fix

uuuvn:


CC-Yeh:
How about version `26.0` in bindings?

uuuvn:
> How about version `26.0` in bindings?

?

CC-Yeh:
> > How about version `26.0` in bindings?
> 
> ?

nvm, I thought it's related

uuuvn:
We use mpp apis that are only available since 26.4, but had our min os version incorrectly declare 26.0 which was the source of the bug, xcode 26.5 checks that and switches implementations based on < or >= 26.2 for whatever reason

PR #409, opened 2026-05-14
dsl buffer args

agolokoz:
Instead of `DenseBuffer` kernels took `Buffer` as buffer arguments


crates/backend-uzu/build/common/traitgen.rs
```diff
@@ -64,15 +62,30 @@ pub fn traitgen(kernel: &Kernel) -> (TokenStream, TokenStream) {
...
+    };
 
     let kernel_trait = quote! {
         pub trait #trait_name: Sized {
             type Backend: crate::backends::common::Backend<Kernels: Kernels<#trait_name = Self>>;
 
             fn new(context: &<Self::Backend as crate::backends::common::Backend>::Context #(, #params)*) -> Result<Self, <Self::Backend as crate::backends::common::Backend>::Error>;
 
-            fn encode<#(#encode_generics, )* 'encoder>(&self, #(#args ,)* encoder: &'encoder mut crate::backends::common::Encoder<Self::Backend>);
+            fn encode<#(#encode_generics),*>(&self, #(#args),*)#maybe_where_block;
```
uuuvn, line 92: Why is this one generic per whole encode instead of per buffer argument? This means that a single kernel won't be able to take eg one dense and one sparse buffer in different arguments which eg attention kernels would have to do

crates/backend-uzu/src/backends/cpu/backend.rs
```diff
@@ -1,13 +1,24 @@
...
+    pub fn buffer_downcast<B: Buffer<Backend = Self>>(buffer: &B) -> &UnsafeCell<Pin<Box<[u8]>>> {
+        let buffer = buffer as &dyn Any;
+        if let Some(buffer) = buffer.downcast_ref::<<Self as Backend>::DenseBuffer>() {
+            buffer
+        } else {
+            unreachable!("Unsupported Cpu buffer type")
+        }
+    }
+}
+
```
uuuvn: This should definitely not be impl Cpu, Cpu is a zero-sized marker struct, it shouldn't do this. I think an extension trait on Buffer<Device = Cpu> is the best way to do this.

crates/backend-uzu/src/backends/metal/backend.rs
```diff
@@ -1,12 +1,30 @@
...
+        let buffer = buffer as &dyn Any;
+        if let Some(buffer) = buffer.downcast_ref::<<Self as Backend>::DenseBuffer>() {
+            buffer
+        } else if let Some(buffer) = buffer.downcast_ref::<MetalSparseBuffer>() {
+            buffer.mtl_buffer()
+        } else {
+            unreachable!("Unsupported Metal buffer type")
+        }
+    }
+}
```
uuuvn: This should definitely not be impl Metal, Metal is a zero-sized marker struct, it shouldn't do this. I think an extension trait on Buffer<Device = Metal> is the best way to do this.

uuuvn (changes requested):

uuuvn (approved):

PR #419, opened 2026-05-15
Support post_layer_scalar

CC-Yeh:


crates/backend-uzu/src/encodable_block/layer/executables.rs
```diff
@@ -47,6 +49,29 @@ impl<B: Backend> LayerExecutables<B> {
...
+                "layer {layer_index} sets post_layer_scalar but has no post_mlp_norm"
+            );
+            let leaf = decoder_layer_loader.leaf("post_layer_scalar").expect("Failed to read post_layer_scalar weight");
+            let allocation = leaf.read_allocation().expect("Failed to read post_layer_scalar weight");
+            let scalar = match leaf.data_type() {
+                DataType::BF16 => Some(allocation_to_vec::<B, bf16>(&allocation)[0].to_f32()),
+                DataType::F16 => Some(allocation_to_vec::<B, f16>(&allocation)[0].to_f32()),
+                DataType::F32 => Some(allocation_to_vec::<B, f32>(&allocation)[0]),
+                _ => None,
+            }
```
uuuvn: Can you add a weights loader function to load into vec directly instead of doing this?

crates/backend-uzu/src/encodable_block/layer/executables.rs
```diff
@@ -47,6 +49,29 @@ impl<B: Backend> LayerExecutables<B> {
...
+            .expect("post_layer_scalar must be a float dtype");
+            Some(scalar)
+        } else {
+            None
+        };
+        let (residual_sum_scalar, output_scalar) = match post_layer_scalar {
+            Some(scalar) => (PostLayerScalar::ScaleResidualSum(scalar), PostLayerScalar::ScaleOutput(scalar)),
+            None => (PostLayerScalar::None, PostLayerScalar::None),
+        };
```
uuuvn: Why is this match separate from the if right before it?

uuuvn (changes requested):

uuuvn (approved):

PR #420, opened 2026-05-15
Support normalize values

CC-Yeh:


eugenebokhan (approved):

uuuvn:
Where did you find `value_norm_config` in a config?

uuuvn (changes requested):
^

crates/backend-uzu/src/config/attention.rs
```diff
@@ -23,5 +25,23 @@ pub struct AttentionConfig {
...
+        let base = self.query_norm_config.as_ref().or(self.key_norm_config.as_ref());
+        Some(NormalizationConfig {
+            scale_precision: base.map_or(ConfigDataType::Float32, |norm| norm.scale_precision),
+            accumulation_precision: base.map_or(ConfigDataType::Float32, |norm| norm.accumulation_precision),
+            epsilon: VALUE_NORM_EPSILON,
+            scale_offset: None,
+            upcast_mode: base.map_or(UpcastMode::FullLayer, |norm| norm.upcast_mode.clone()),
+            subtract_mean: false,
+            use_bias: false,
+        })
```
uuuvn: ??? Defaulting ConfigDataType::Float32/UpcastMode::FullLayer/anything else like that in general anywhere is a very bad idea, it's just plain wrong and will silently do the wrong thing if ever triggered, we should never do stuff like this. Even panic is preferred to silently doing something wrong as otherwise that will make us waste hours on debugging if we're lucky or just degrade model intelligence a little bit if we're not and we'll never know that models running on uzu are a bit dumber than reference.

We're normalizing output of a qkv projection and so its activation_precision should always be used here regardless of what q or k projection has, upcast mode should (probably? I don't think it matters when scale is 1, but not sure) be FullLayer (again regardless of what q or k projection is) as that matches what lalamo does:

```python
def _rms_normalize(
    x: Float[Array, "*batch channels"],
    eps: float,
) -> Float[Array, "*batch channels"]:
    variance = jnp.mean(jnp.square(x.astype(jnp.float32)), axis=-1, keepdims=True) + eps
    return (x * jax.lax.rsqrt(variance)).astype(x.dtype)
```

```python
if self.config.normalize_values:
    values = _rms_normalize(values, eps=1e-6)
```

uuuvn (changes requested):

crates/backend-uzu/src/config/attention.rs
uuuvn: I don't think this particular case is at all likely to actually fire (requires a combination of lack of both q/k norm and presence of v + non-fp32 activation precision or q/k norm somehow having a different activation datatype from qkv projection which basically can't happen) but it's a general very harmful pattern that we should never do in uzu, especially when there is a much more correct thing to do (activation precision from qkv, always full layer, always fp32 acc, that matches lalamo exactly)

CC-Yeh: Thanks for catching that!

uuuvn (approved):

PR #427, opened 2026-05-16
Kv cache layer sparse

agolokoz:
Use full length sparse buffers for kv cache layers

### DIFF
--- a/crates/backend-uzu/src/backends/common/context.rs
+++ b/crates/backend-uzu/src/backends/common/context.rs
@@ -43,6 +43,12 @@ pub trait Context: Sized {
         capacity: usize,
     ) -> Result<<Self::Backend as Backend>::SparseBuffer, <Self::Backend as Backend>::Error>;
 
+    /// Block until every sparse map/unmap operation issued so far has been
+    /// committed on the GPU. Must be called before any kernel on the main
+    /// command queue can read or write pages that were just mapped, because
+    /// sparse mapping ops run on a separate (MTL4) queue.
+    fn wait_for_pending_sparse_mappings(&self) -> Result<(), <Self::Backend as Backend>::Error>;
+
     fn peak_memory_usage(&self) -> Option<usize>;
 
     fn enable_capture();
--- a/crates/backend-uzu/src/backends/common/context.rs
+++ b/crates/backend-uzu/src/backends/common/context.rs
@@ -43,10 +43,9 @@ pub trait Context: Sized {
         capacity: usize,
     ) -> Result<<Self::Backend as Backend>::SparseBuffer, <Self::Backend as Backend>::Error>;
 
-    /// Block until every sparse map/unmap operation issued so far has been
-    /// committed on the GPU. Must be called before any kernel on the main
-    /// command queue can read or write pages that were just mapped, because
-    /// sparse mapping ops run on a separate (MTL4) queue.
+    /// Block until every sparse map/unmap operation issued so far has been committed on the GPU.
+    /// Must be called before any kernel on the main command queue can read or write pages that were just mapped,
+    /// because sparse mapping ops run on a separate (MTL4) queue.
     fn wait_for_pending_sparse_mappings(&self) -> Result<(), <Self::Backend as Backend>::Error>;
 
     fn peak_memory_usage(&self) -> Option<usize>;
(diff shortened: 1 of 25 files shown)


crates/backend-uzu/src/backends/common/kernel/attention.rs
```diff
@@ -56,11 +56,17 @@ impl<B: Backend> AttentionGemmBlock<B> {
...
-        args: AttentionGemmArguments<B>,
-    ) -> Result<(), B::Error> {
+        args: AttentionGemmArguments<B, Keys, Values>,
+    ) -> Result<(), B::Error>
+    where
+        Keys: AsBufferRangeRef,
+        Keys::Buffer: Buffer<Backend = B>,
+        Values: AsBufferRangeRef,
+        Values::Buffer: Buffer<Backend = B>,
+    {
```
uuuvn: Does `Keys: AsBufferRangeRef<Buffer: Buffer<Backend = B>>` or something like that work here? In general we should avoid big ugly where clauses wherever possible

crates/backend-uzu/src/backends/common/kernel/matmul.rs
```diff
@@ -53,9 +53,11 @@ pub trait MatmulKernel: Sized {
...
+    fn encode<TB>(
         &mut self,
-        arguments: MatmulArguments<Self::Backend>,
+        arguments: MatmulArguments<Self::Backend, TB>,
         encoder: &mut Encoder<Self::Backend>,
-    );
+    ) where
+        TB: AsBufferRangeRef,
+        TB::Buffer: Buffer<Backend = Self::Backend>;
 }
```
uuuvn, line 61: Same here (and everywhere in general) with where clauses. They're much harder to read than inline bounds because generic declaration is separate from the bound and the bound itself consists of (in this case) of two separate bounds. With `TB: AsBufferRangeRef<Buffer: Buffer<Backend = Self::Backend>` it's much easier to read

crates/backend-uzu/src/backends/common/context.rs
```diff
@@ -43,6 +43,11 @@ pub trait Context: Sized {
         capacity: usize,
     ) -> Result<<Self::Backend as Backend>::SparseBuffer, <Self::Backend as Backend>::Error>;
 
+    /// Block until every sparse map/unmap operation issued so far has been committed on the GPU.
+    /// Must be called before any kernel on the main command queue can read or write pages that were just mapped,
+    /// because sparse mapping ops run on a separate (MTL4) queue.
+    fn wait_for_pending_sparse_mappings(&self) -> Result<(), <Self::Backend as Backend>::Error>;
```
uuuvn, line 49: This is not a good way to do it, separate MTL4 queue is an implementation detail of metal and it shouldn't leak into common code if it can be efficiently done inside of the backend which here it can by eg using MTLEvent and having context add signals and waits around command buffer submissions.

The most obvious way to do it is timeline_event: MTLEvent and timeline_value: RefCell<u64>, when we submit anything to either queue we actually submit wait for current timeline_value on timeline_event before doing work, then bump timeline_value (+1), signal new value after work. This should also get rid of the AsyncBuffers.event and the whole public event api if I'm not missing anything

crates/backend-uzu/src/forward_pass/kv_cache_layer.rs
```diff
@@ -44,14 +44,30 @@ pub const INVALID_POSITION: usize = i32::MAX as usize;
 pub struct KVCacheLayer<B: Backend> {
     pub state: KVCacheLayerState,
     /// [max_prefix_length + max_suffix_length, num_groups, head_dim]
-    pub keys: Allocation<B>,
+    pub keys: B::SparseBuffer,
     /// [max_prefix_length + max_suffix_length, num_groups, head_dim]
-    pub values: Allocation<B>,
+    pub values: B::SparseBuffer,
```
uuuvn: We want to have a non-sparse option for backends that don't support it (cpu, webgpu). I think the right shape is `KVCacheLayer<B: Backend, Buf: Buffer<Backend = B>>`

uuuvn (changes requested):
I see this is work in progress, but left some preliminary reviews to make things faster

PR #430, opened 2026-05-18
Sync fix

agolokoz:


crates/backend-uzu/src/backends/common/context.rs
```diff
@@ -62,4 +57,7 @@ pub trait Context: Sized {
     fn tf32_enabled() -> bool {
         env::var("UZU_TF32").map(|v| v == "1" || v.eq_ignore_ascii_case("true")).unwrap_or(false)
     }
+
+    /// Call after batch of map/unmap operations of sparse buffers
+    fn sparse_mappings_signal(&self);
```
agolokoz, line 62: This is necessary in common code, because only there we know when all mapping operations are submitted

PR #432, opened 2026-05-19
Unified Quantized Gemm

eugenebokhan:
| Shape | Baseline | Unified | Delta |
|---|---:|---:|---:|
| sb/bf16/gs128/M[5]N[14336]K[4096] | 484.4 µs | 533.1 µs | +10.1% |
| sb/bf16/gs128/M[8]N[14336]K[4096] | 533.9 µs | 588.5 µs | +10.2% |
| sb/bf16/gs128/M[16]N[14336]K[4096] | 1440.9 µs | 1306.4 µs | −9.3% |
| sb/bf16/gs128/M[24]N[14336]K[4096] | 2238.3 µs | 2119.1 µs | −5.3% |
| sb/bf16/gs128/M[32]N[14336]K[4096] | 2799.0 µs | 955.6 µs | **−65.9%** |
| sb/bf16/gs128/M[40]N[14336]K[4096] | 3242.6 µs | 2306.7 µs | −28.9% |
| sb/bf16/gs128/M[48]N[14336]K[4096] | 2431.6 µs | 2346.0 µs | −3.5% |
| sb/bf16/gs128/M[64]N[14336]K[4096] | 2569.7 µs | 2386.6 µs | −7.1% |
| sb/bf16/gs128/M[128]N[14336]K[4096] | 4731.1 µs | 4391.3 µs | −7.2% |
| sb/bf16/gs128/M[32]N[4096]K[4096] | 632.8 µs | 279.0 µs | **−55.9%** |
| sb/bf16/gs128/M[64]N[4096]K[4096] | 633.6 µs | 586.0 µs | −7.5% |
| sb/bf16/gs128/M[32]N[4096]K[14336] | 2719.7 µs | 1096.4 µs | **−59.7%** |
| sb/bf16/gs128/M[64]N[4096]K[14336] | 2397.2 µs | 2028.8 µs | −15.4% |
| sb/bf16/gs64/M[32]N[14336]K[4096] | 2497.1 µs | 945.3 µs | **−62.1%** |
| sb/bf16/gs64/M[64]N[14336]K[4096] | 2256.1 µs | 2011.0 µs | −10.9% |
| zp/bf16/gs128/M[16]N[14336]K[4096] | 1128.5 µs | 1150.0 µs | +1.9% |
| zp/bf16/gs128/M[32]N[14336]K[4096] | 2695.4 µs | 1000.1 µs | **−62.9%** |
| zp/bf16/gs128/M[48]N[14336]K[4096] | 2154.0 µs | 2046.7 µs | −5.0% |
| zp/bf16/gs128/M[64]N[14336]K[4096] | 2049.0 µs | 1899.5 µs | −7.3% |
| zp/bf16/gs64/M[32]N[14336]K[4096] | 1674.2 µs | 958.3 µs | −42.8% |
| zp/bf16/gs64/M[64]N[14336]K[4096] | 1887.4 µs | 1676.9 µs | −11.2% |
| zp/f16/gs64/M[32]N[14336]K[4096] | 764.2 µs | 482.2 µs | −36.9% |
| zp/f16/gs64/M[64]N[14336]K[4096] | 922.4 µs | 935.2 µs | +1.4% |


eugenebokhan:
Screenshot 2026-05-24 at 23 37 43

crates/backend-uzu/src/backends/metal/kernel/matmul/gemm/kernel.rs
```diff
@@ -42,74 +51,285 @@ impl GemmKernel {
...
         }
     }
 
-    pub(crate) fn encode<T: AsBufferRangeRef<Buffer: Buffer<Backend = Metal>>>(
+    pub(crate) fn encode<'a, TB: AsBufferRangeRef<Buffer: Buffer<Backend = Metal>>>(
         &mut self,
         context: &MetalContext,
-        dispatch: GemmDispatch<'_, Metal, T>,
+        arguments: MatmulArguments<'a, Metal, TB>,
+        force_simdgroup: bool,
```
CC-Yeh: How about

```
 pub enum MxuMode { Auto, ForceOn, ForceOff }
```

I think it's cleaner and more explicit with its intention

crates/backend-uzu/src/backends/cpu/kernel/matmul/matmul.rs
```diff
@@ -134,21 +170,148 @@ impl MatmulKernel for MatmulCpuKernel {
...
+                ..
+            } => unreachable!(),
+        };
+        if !matches!(group_size, 32 | 64 | 128) {
+            return Err(MatmulError::UnsupportedGroupSize(group_size as usize));
+        }
+
+        if arguments.m >= 5 && arguments.n > 1 {
+            encode_quantized_gemm(encoder, arguments, self.data_type);
+            return Ok(());
```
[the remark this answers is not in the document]
CC-Yeh: @eugenebokhan I think this comment make sense.

crates/backend-uzu/src/backends/cpu/kernel/matmul/matmul.rs
```diff
@@ -134,21 +170,148 @@ impl MatmulKernel for MatmulCpuKernel {
...
+            return Err(MatmulError::UnsupportedDOp {
+                bit: GemmDTransform::ACCUMULATE,
+                path: "MatmulCpuKernel/Quant",
+            });
+        }
+        if d_mask.contains(GemmDTransform::BIAS) {
+            return Err(MatmulError::UnsupportedDOp {
+                bit: GemmDTransform::BIAS,
+                path: "MatmulCpuKernel/Quant",
+            });
```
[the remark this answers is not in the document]
CC-Yeh: @eugenebokhan I think this one also make sense.

crates/backend-uzu/src/encodable_block/linear/matmul.rs
```diff
@@ -0,0 +1,360 @@
...
+                    b: &self.weights,
+                    scales,
+                    zero_points: zero_points_or_biases,
+                    mode: *mode,
+                    group_size: *group_size,
+                },
+            },
+        };
+
+        let mut d_transform: HashSet<MatmulDOp<'_, B>> = HashSet::new();
```
CC-Yeh: Do we really need `HashSet`?

CC-Yeh: There will be extra allocations + lookups in `encode` from it

crates/backend-uzu/src/backends/common/kernel/matmul.rs
```diff
@@ -1,48 +1,154 @@
...
+    ScaleZeroPointDequant {
+        b: &'a Allocation<B>,
+        scales: &'a Allocation<B>,
+        zero_points: &'a Allocation<B>,
+        mode: QuantizationMode,
+        group_size: u32,
+    },
+}
+
+pub enum MatmulDOp<'a, B: Backend> {
```
CC-Yeh: The `HashSet<MatmulDOp>` feels off here.

The discriminant-only Hash/PartialEq means `Bias{a} == Bias{b}`, so duplicate inserts silently drop the second value. The "at most one of each kind" invariant lives in container semantics rather than the type.

CC-Yeh: Worth considering a plain struct with Option<&Allocation> fields + a bits() helper for the bitmask. Same expressive power, no allocation in the inference hot path, and the at-most-one invariant becomes unforgeable.

crates/backend-uzu/src/backends/metal/kernel/matmul/mod.rs
```diff
@@ -35,46 +43,299 @@ fn max_gemv_batch_threshold() -> u32 {
...
+            k,
+        } = arguments;
+        d_transform.retain(|op| op.bit() != GemmDTransform::BIAS);
+
+        self.quant_gemv.encode(
+            encoder,
+            MatmulArguments {
+                a,
+                a_offset,
+
```
CC-Yeh:
```suggestion
```

crates/backend-uzu/src/backends/metal/kernel/matmul/mod.rs
```diff
@@ -35,46 +43,299 @@ fn max_gemv_batch_threshold() -> u32 {
...
+                d_transform,
+                m,
+                n,
+                k,
             },
+        )?;
+
+        if let Some(bias) = post_bias {
+            self.bias_add.encode(None::<&<Metal as Backend>::DenseBuffer>, bias, d, n, m * n, encoder);
+        }
```
CC-Yeh: Can we consider fusing bias into the GEMV epilogue? This runs every linear every token on decode, and `dispatch_quant_gemm` already fuses it.

crates/backend-uzu/src/backends/common/kernel/matmul/mod.rs
```diff
@@ -1,48 +1,154 @@
...
+        &self,
+        other: &Self,
+    ) -> bool {
+        discriminant(self) == discriminant(other)
+    }
+}
+
+impl<B: Backend> Eq for MatmulDOp<'_, B> {}
+
 pub struct MatmulArguments<'a, B: Backend, TB: AsBufferRangeRef = Allocation<B>> {
```
CC-Yeh: The three `b_*` layout fields are only meaningful for `FullPrecision`, the quant variants reject them via runtime guards in `dispatch_quant_gemv`, `dispatch_quant_gemm`, and CPU encode_quant, and the CPU guard is missing the `b_offset != 0` check the Metal paths have (silent footgun).

Moving these into the `FullPrecision` variant makes the invariants unrepresentable and deletes the three guards. Same shape problem as `MatmulDOp` — the type permits states the runtime then rejects

CC-Yeh (changes requested):

crates/backend-uzu/src/backends/metal/kernel/matmul/gemm/specialization.rs
```diff
@@ -0,0 +1,144 @@
...
+                return Err(GemmSpecializationError::QuantizedRequiresSimdgroup);
+            }
+            if !self.transpose_b {
+                return Err(GemmSpecializationError::QuantizedRequiresTransposedB);
+            }
+        }
+        Ok(())
+    }
+
+    pub(crate) fn precompile_configs(data_type: DataType) -> Vec<Self> {
```
CC-Yeh, line 47: I think most PSOs are never used, should we just lazily compile those at first run?

crates/backend-uzu/src/backends/metal/kernel/matmul/gemm/kernel.rs
eugenebokhan: What should ForceOn do on the old chips?

crates/backend-uzu/src/encodable_block/linear/matmul.rs
eugenebokhan: I don't like the hash set runtime overhead but at the same time would like to keep the compile time gating of improper use of the ops

crates/backend-uzu/src/backends/common/kernel/matmul.rs
eugenebokhan: I thought about this approach but it can cause improper use of the ops as well, because there might be a mismatch of the bitset with the associated data

crates/backend-uzu/src/backends/metal/kernel/matmul/mod.rs
eugenebokhan: I explicitly kept the GEMV untouched to focus on GEMM in this PR. I'm planning to extrapolate the fusion to the gemv in the following PR

crates/backend-uzu/src/backends/common/kernel/matmul/mod.rs
eugenebokhan: This is done intentionally, currently I'm in the process of scaffolding the API while will adding the impl gaps in the upcoming PRs.

crates/backend-uzu/src/backends/metal/kernel/matmul/gemm/kernel.rs
CC-Yeh: [nit]

Then maybe just

```
pub enum MxuMode { Auto, ForceOff }
```

crates/backend-uzu/src/backends/common/kernel/matmul.rs
CC-Yeh: Think we're picturing different structs. If it's

```rust
  struct MatmulDTransform { bits: GemmDTransform, bias: Option<...>, ... }
```
  then yeah, bits and bias can disagree.

  What I had in mind has no bits field, just payloads — `bits()` is a method:

```
  struct MatmulDTransform {
      ab_scale: f32,
      accumulate: bool,
      bias: Option<&Allocation<B>>,
      rht_factors: Option<&Allocation<B>>,
  }

  impl MatmulDTransform {
      fn bits(&self) -> GemmDTransform {
          let mut m = empty();
          if self.bias.is_some()        { m |= BIAS; }
          if self.rht_factors.is_some() { m |= RHT; }
          if self.ab_scale != 1.0       { m |= SCALE; }
          if self.accumulate            { m |= ACCUMULATE; }
          m
      }
  }
  ```
  
 Nothing to desync — fields are the source of truth, mask is derived. Does that work?

crates/backend-uzu/src/backends/metal/kernel/matmul/gemm/kernel.rs
eugenebokhan: Then it is just simply a bool, right?

crates/backend-uzu/src/backends/metal/kernel/matmul/gemm/specialization.rs
eugenebokhan: The benchmarks can have a warmup stage, but the real world usage experience may degrade in this case

crates/backend-uzu/src/backends/metal/kernel/matmul/gemm/kernel.rs
CC-Yeh: Fair, just first look at `_simdgroup` is a bit confusing for me or for anyone doesn't know the details of GEMM implementation

eugenebokhan:
What remains unimplemented and will be added later: RHT fusion as a D transform. MXU support to the quantized path.

crates/backend-uzu/src/backends/metal/kernel/matmul/gemm/specialization.rs
CC-Yeh: Pruning isn't full lazy, the shape axes (tiling, alignment, output_transform) stay eager, so first inference is just as hot as today. Only group_size/bits/prologue get dropped, and those PSOs can't be dispatched to anyway given the model's quant config.

For a 32-layer model that's roughly 80k get_or_create calls → 9k at load, so startup actually gets faster. First token and steady-state are unchanged. Also saves ~25MB (GPU PSO bytes + the per-linear HashMaps), which is small but nice on phones.

crates/backend-uzu/src/backends/metal/kernel/matmul/gemm/kernel.rs
eugenebokhan: I ended up using the encode_dispatch_path that eliminates the force_simdgroup bool

crates/backend-uzu/src/backends/metal/kernel/matmul/gemm/specialization.rs
eugenebokhan: Done

crates/backend-uzu/src/backends/common/kernel/matmul.rs
eugenebokhan: Done

crates/backend-uzu/src/encodable_block/linear/matmul.rs
eugenebokhan: Done

uuuvn (approved):

PR #435, opened 2026-05-20
New config dsl and lalamo 0.10.0 config format

uuuvn:


uuuvn:
(I've set this as ready to review for codex, humans please disregard for now)

PR #436, opened 2026-05-21
Add checking if sparse buffers supported for Metal

agolokoz:


Cargo.toml
```diff
@@ -69,7 +69,7 @@ tokenizers = { version = "0.22", features = [
 objc2 = "0.6.4"
 objc2-foundation = "0.3.2"
 objc2-security = "0.3.2"
-metal = { package = "mtl-rs", version = "0.1" }
+metal = { package = "mtl-rs", version = "0.1.12" }
```
eugenebokhan, line 72: no need to set patch version, simply use cargo update to update the lock file to the latest one

PR #437, opened 2026-05-22
Fix Linux CPU-only download-manager build

KonstantinSKY:
## What

Fix Linux CPU-only builds by compiling the Apple download backend only on Apple targets.

## Why

`download-manager` exposed Apple-specific modules/imports on Linux even though Apple dependencies are target-gated. This breaks CPU-only Linux checks.

## Validation

- `cargo check -p uzu --no-default-features --features backend-cpu`
- `cargo check -p cli --no-default-features --features backend-cpu,capability-cli`
- `cargo test -p backend-uzu --no-default-features --lib`
- `cargo run --release -p uzu --example quick_start --no-default-features --features backend-cpu`

### DIFF
--- a/crates/download-manager/src/backends/common/backend.rs
+++ b/crates/download-manager/src/backends/common/backend.rs
@@ -10,6 +10,7 @@ use crate::{
 
 pub enum InitialTaskAttachment<B: Backend> {
     None,
+    #[cfg_attr(not(target_vendor = "apple"), allow(dead_code))]
     Downloading {
         active_task: B::ActiveTask,
         initial_downloaded_bytes: u64,
--- a/crates/download-manager/src/backends/mod.rs
+++ b/crates/download-manager/src/backends/mod.rs
@@ -1,3 +1,4 @@
+#[cfg(target_vendor = "apple")]
 pub mod apple;
 pub mod common;
 pub mod universal;
--- a/crates/download-manager/src/file_download_manager.rs
+++ b/crates/download-manager/src/file_download_manager.rs
@@ -5,8 +5,10 @@ use tokio_stream::wrappers::BroadcastStream as TokioBroadcastStream;
 
 use crate::{
     DownloadError, DownloadId, FileCheck, FileDownloadEvent, FileDownloadTask,
-    backends::{apple::AppleDownloadManager, universal::UniversalDownloadManager},
+    backends::universal::UniversalDownloadManager,
 };
+#[cfg(target_vendor = "apple")]
+use crate::backends::apple::AppleDownloadManager;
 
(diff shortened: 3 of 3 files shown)


crates/download-manager/src/backends/common/backend.rs
```diff
@@ -10,6 +10,7 @@ use crate::{
 
 pub enum InitialTaskAttachment<B: Backend> {
     None,
+    #[cfg_attr(not(target_vendor = "apple"), allow(dead_code))]
```
uuuvn, line 13: Why is this `allow(dead_code)`?

uuuvn (changes requested):

crates/download-manager/src/backends/common/backend.rs
KonstantinSKY: The cfg_attr is silencing a dead_code warning that becomes structurally unavoidable once pub mod apple; is cfg-gated.

  InitialTaskAttachment::Downloading is constructed only in backends/apple/backend.rs at line 69, where it wraps an NSURLSessionDownloadTask via B::ActiveTask. Once the apple submodule is gated on target_vendor
  = "apple", that constructor disappears on non-Apple targets, and the variant is never created there.

  The variant has to stay in the type, though, because shared code in file_download_task_actor/generic_file_download_task.rs pattern-matches on it at lines 99 and 105 to drive the download lifecycle. Removing
  the variant would break that match.

  Result: rustc emits "variant 'Downloading' is never constructed" on non-Apple targets. And because CARGO_ENCODED_RUSTFLAGS is set to "-Dwarnings" in tests.yml at line 79, that warning fails CI on Linux and
  wasm32.

  I looked at two alternatives before going with cfg_attr:

  1. Cfg-gate the variant itself, with #[cfg(target_vendor = "apple")] Downloading { ... }. That would make the variant disappear on non-Apple, but then the two match arms in generic_file_download_task.rs also
  need cfg-gating (or a wildcard fallback). The shared download lifecycle ends up splitting by target, which is more diff and harder to follow.

  2. Replace the match arms with _ => .... Smaller diff, but loses the exhaustiveness check on Apple. If a future variant is added, the wildcard silently swallows it.

  The chosen cfg_attr keeps the type identical across all targets, leaves the shared matching code untouched, and only suppresses the warning where the constructor is genuinely absent.

  I can switch to the cfg-gate-variant approach if you'd rather have that trade-off, just thought the surgical attribute was the smaller change.

uuuvn:
I assume this is not a draft anymore?

uuuvn (approved):

KonstantinSKY:
Thanks for reviewing and merging this. Yes, it was ready — I just missed marking the PR as ready for review.

PR #442, opened 2026-05-24
ci: add Linux CPU build check (#437 follow-up)

KonstantinSKY:
## What

Add a Linux CPU-only entry to the existing `tests` matrix as a follow-up to #437.

## Why

#437 fixed the Linux CPU-only build by gating the Apple download backend behind `target_vendor = "apple"`, but CI did not build a non-Apple CPU-only configuration, so the fix could silently regress.

## Validation

The Linux matrix entry runs on `ubuntu-latest` under `-Dwarnings` and checks:

- `cargo +stable check -p backend-uzu --no-default-features`
- `cargo +stable check -p uzu --no-default-features --features backend-cpu`
- `cargo +stable check -p cli --no-default-features --features backend-cpu,capability-cli`

Verified locally on x86_64-linux:

- `/tmp/actionlint -ignore 'label "mac" is unknown' .github/workflows/tests.yml`
- `cargo +stable check -p backend-uzu --no-default-features`
- `cargo +stable check -p uzu --no-default-features --features backend-cpu`
- `cargo +stable check -p cli --no-default-features --features backend-cpu,capability-cli`


.github/workflows/tests.yml
```diff
@@ -142,14 +142,38 @@ jobs:
...
+        run: sudo apt-get update && sudo apt-get install -y libasound2-dev
+
+      - name: Check backend-uzu (CPU-only, no default features)
+        run: cargo +stable check -p backend-uzu --no-default-features
+
+      - name: Check uzu (CPU-only)
+        run: cargo +stable check -p uzu --no-default-features --features backend-cpu
+
+      - name: Check cli (CPU-only)
+        run: cargo +stable check -p cli --no-default-features --features backend-cpu,capability-cli
```
uuuvn: Why do we need a separate job for linux cpu check? We already have a tests matrix, it should just be another entry in there

uuuvn (changes requested):

.github/workflows/tests.yml
KonstantinSKY: Updated. I kept it separate initially because it needed a different runner and Linux package setup, and I wanted to avoid changing the existing tests matrix behavior. But I agree this fits better as a matrix
  entry, so I folded it into `tests` and removed the separate job. Thank you.

.github/workflows/tests.yml
```diff
@@ -55,22 +55,33 @@ jobs:
...
-    runs-on: [ self-hosted, macos, mac, arm64 ]
+    runs-on: ${{ fromJSON(matrix.runs_on) }}
 
     strategy:
       matrix:
         backend: [ 'cpu', 'metal' ]
         target: [ '' ]
         tracing: [ false, true ]
+        runs_on: [ '["self-hosted","macos","mac","arm64"]' ]
+        linux_cpu: [ false ]
```
uuuvn: Why are runs_on and linux_cpu separate things?

Let's have one host: macos/linux and both are derived from it in job.runs-on and other places with host-dependent behavior

.github/workflows/tests.yml
```diff
@@ -87,15 +98,31 @@ jobs:
...
           test_data: true
           nextest: true
           target: ${{ matrix.target }}
 
+      - uses: dtolnay/rust-toolchain@stable
+        if: ${{ matrix.linux_cpu }}
+
+      - name: Install Linux CPU dependencies
+        if: ${{ matrix.linux_cpu }}
+        run: sudo apt-get update && sudo apt-get install -y libasound2-dev
```
uuuvn: Why are you gating setup on non-linux and then copy-pasting part of setup? This is what setup inputs are for

.github/workflows/tests.yml
```diff
@@ -87,15 +98,31 @@ jobs:
...
       - name: Build crates
+        if: ${{ !matrix.linux_cpu }}
         run: cargo +stable build --no-default-features $FEATURE_FLAGS ${{ matrix.target == 'wasm32-wasip1-threads' && '-p uzu' || '' }}
 
+      - name: Check Linux CPU crates
+        if: ${{ matrix.linux_cpu }}
+        run: |
+          cargo +stable check -p backend-uzu --no-default-features
+          cargo +stable check -p uzu --no-default-features --features backend-cpu
+          cargo +stable check -p cli --no-default-features --features backend-cpu,capability-cli
```
uuuvn: Why are you adding a separate step for "cpu crates"? There is no such thing as "cpu crates". Do all crates not build on linux cpu? If so the two reasonable things are either fix the build for them or add a similar -p as with wasm

uuuvn (changes requested):

KonstantinSKY:
Updated, thanks for the review.

- Collapsed the parallel `runs_on`/`linux_cpu` fields into a single `host` axis (`macos`/`linux`); `runs-on`, the job name, and the host-specific setup all derive from it.
- Stopped gating `setup` off on Linux — it now runs everywhere and installs the toolchain, and the only Linux-specific dependency (`libasound2-dev`) is handled by a new `linux` input on the setup action.
- Dropped the separate "cpu crates" step. The Linux row now reuses the existing `Build crates` step, narrowed with `-p cli --features backend-cpu,capability-cli`, the same way the wasm row uses `-p uzu`. `cli` transitively builds `uzu`, `backend-uzu` and `download-manager`, so the #437 CPU-only path stays covered.

Verified on x86_64-linux with `-Dwarnings`:
`cargo +stable build --no-default-features -p cli --features backend-cpu,capability-cli`

uuuvn (approved):

KonstantinSKY:
thank you!

PR #443, opened 2026-05-25
Support Per Layer Embedding

CC-Yeh:


CC-Yeh:
Will fuse some small kernels in PLE (it's very small in gemma4, should not matter much in performance) after supporting gemma4

PR #444, opened 2026-05-26
Add Vulkan backend scaffold

KonstantinSKY:
## What

Add the initial cfg-gated Vulkan backend scaffold under `backend-uzu`.

## Why

This creates a small, compile-checked entry point for follow-up Vulkan backend work without changing existing CPU or Metal behavior.

## Validation

- `cargo +stable check -p backend-uzu --no-default-features`
- `cargo +stable check -p backend-uzu --no-default-features --features vulkan`
- `env CARGO_ENCODED_RUSTFLAGS=-Dwarnings cargo +stable check -p backend-uzu --no-default-features --features vulkan`
- `cargo metadata --locked --format-version=1`
- `cargo +nightly fmt --check`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -221,6 +221,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "ash"
+version = "0.38.0+1.3.281"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0bb44936d800fea8f016d7f2311c6a4f97aebd5dc86f09906139ec848cf3a46f"
+dependencies = [
+ "libloading 0.8.9",
+]
+
+[[package]]
 name = "askama"
 version = "0.14.0"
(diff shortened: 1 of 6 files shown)


uuuvn:
This "scaffold" serves no purpose in main. We have an old [vulkan](https://github.com/trymirai/uzu/tree/vulkan) branch, but our plans regarding other backends have changed to first do [webgpu](https://github.com/trymirai/uzu/tree/webgpu) and then reuse it's infrastructure for vulkan later

KonstantinSKY:
Thanks for clarifying, understood. I won’t open more standalone Vulkan PRs against `main`.

For the WebGPU-first direction, is the current `webgpu` branch / #380 still the right public base for external PRs, or are you planning a new WebGPU branch from the current `main`? If there will be a new base, I can target small follow-up PRs there instead, so I don’t add work to the older dev branch.

PR #446, opened 2026-05-26
Unified GEMV

eugenebokhan:
[compare-a18-pro.txt](https://github.com/user-attachments/files/28464030/compare-a18-pro.txt)
[compare-m2-pro.txt](https://github.com/user-attachments/files/28464023/compare-m2-pro.txt)
[compare-m2.txt](https://github.com/user-attachments/files/28464031/compare-m2.txt)
[compare-m4.txt](https://github.com/user-attachments/files/28464034/compare-m4.txt)
[compare-m4-max.txt](https://github.com/user-attachments/files/28464032/compare-m4-max.txt)
[compare-m4-pro.txt](https://github.com/user-attachments/files/28464033/compare-m4-pro.txt)


crates/backend-uzu/src/backends/metal/kernel/matmul/gemv/kernel.rs
```diff
@@ -1,65 +1,257 @@
...
+        };
+        let input_aligned = k % block_size == 0;
+        let (num_simdgroups, k_split) = if is_quant {
+            // Output RHT needs a full 32-row threadgroup (8 simdgroups x 4 rows)
+            // so the simdgroup-wide Hadamard covers one 32-element block.
+            (
+                if rht_factors.is_some() {
+                    8u32
+                } else {
+                    2u32
```
CC-Yeh: why 2 simdgroups? Do you see any performance difference?

crates/backend-uzu/tests/performance/kernel/matmul/quant_gemv_bench.rs
```diff
@@ -1,14 +1,9 @@
-#![cfg(metal_backend)]
-
 use backend_uzu::{
```
CC-Yeh, line 1: Let's remove M=8 (GEMM) from gemv benchmarks

crates/backend-uzu/src/backends/metal/kernel/matmul/gemv/kernel.rs
```diff
@@ -1,65 +1,257 @@
...
+        Ok(Self {
             weights_data_type,
             input_data_type,
             output_data_type,
             pipelines: HashMap::new(),
-        };
-        for &config in GemvSpecialization::precompile_configs(weights_data_type) {
-            kernel.get_or_create(context, config)?;
-        }
-        Ok(kernel)
```
CC-Yeh, line 35: Why no longer precompile? Can we precompile some frequent used shapes?

crates/backend-uzu/src/backends/metal/kernel/matmul/gemv/gemv.metal
```diff
@@ -1,318 +1,432 @@
...
+template <typename WeightT, typename U, GemmBPrologueKind B_PROLOGUE, uint BITS>
+struct QuantRowOffsets {
+  const device WeightT* scales = nullptr;
+  const device WeightT* biases = nullptr;
+  const device uint8_t* zps = nullptr;
+  uint group_stride = 0;
+  uint zp_stride = 0;
+  bool high_nibble = false;
+
+  void load(thread U* scale, thread U* offset) const {
```
CC-Yeh: This looks like only for 4bits/8bits? How does it work with different `BITS` such as 3/5/6?

crates/backend-uzu/src/backends/metal/kernel/matmul/gemv/gemv.metal
```diff
@@ -1,318 +1,432 @@
...
-
-template <typename WeightT, typename InputT, typename OutputT>
+  void advance(uint groups) {
+    scales += groups;
+    if constexpr (B_PROLOGUE == GemmBPrologueKind::ScaleBiasDequant) {
+      biases += groups;
+    } else if constexpr (
+        B_PROLOGUE == GemmBPrologueKind::ScaleZeroPointDequant
+    ) {
+      zps += (BITS == 4) ? (groups / 2) : groups;
```
CC-Yeh: Can we make it read `pack_factor` to be more generic?

crates/backend-uzu/src/backends/metal/kernel/matmul/gemv/gemv.metal
```diff
@@ -1,318 +1,432 @@
...
-// Upper bound for threadgroup memory (in floats).
-// Max config: tg_simd_cols=16, output_rows_per_tg=16, thread_out_rows=4
-// → 16 * (16 + 4) = 320. Unused when tg_simd_cols == 1.
-#define GEMV_MAX_THREADGROUP_MEMORY 320
-
-template <typename WeightT, typename InputT, typename OutputT>
+  void advance(uint groups) {
+    scales += groups;
+    if constexpr (B_PROLOGUE == GemmBPrologueKind::ScaleBiasDequant) {
+      biases += groups;
```
CC-Yeh: I'd like some traits like `has_bias` and `has_zp`, since there are some quantizations use either or both of them, or we could just specialize on `advance` or  `QuantRowOffsets`.

CC-Yeh (commented):
I’d suggest keeping the single merged GEMV entry point, but making the internals
  more Lego-like.

  Right now `gemv.metal` feels like one big function that owns everything: output
  tiling, quant params, K scheduling, tail handling, reduction, bias/RHT/store.
  That makes it hard to review and hard to reason about whether a QMV performance
  change came from tiling, dequant, epilogue, or just register pressure.

  A cleaner direction would be one shared GEMV pipeline made out of small compile-
  time pieces:

  ```metal
  GemvCore<Spec>::run(...) {
    tile = OutputTile<Spec>::make(...);
    b = BSource<Spec>::make(...);

    loop over K:
      x = ALoader<Spec>::load(...);
      BSource<Spec>::accumulate4(b, x, result);

    Reduce<Spec>::run(...);
    Epilogue<Spec>::store(...);
  }

  Where the pieces are named by responsibility:

  - OutputTile: row mapping and tail behavior
  - BSource: full-precision vs quantized B access
  - QuantPrologue: scale-bias / zero-point / symmetric params
  - KSchedule: block size, split-K, aligned/tail K
  - Epilogue: scale, accumulate, bias, RHT, store

  This still preserves the “merged GEMV” design, but the hot QMV path becomes
  inspectable as a composition of small parts instead of being buried in one large
  kernel body.

  On the Rust side, I’d also like the routing to become one selection function:

  GemvSpecialization::select(args, dtypes) -> Option<GemvSpecialization>

  so “is GEMV supported?” and “which GEMV specialization do we instantiate?”
  cannot drift apart.


@eugenebokhan @uuuvn  what do you think?

crates/backend-uzu/tests/performance/kernel/matmul/quant_gemv_bench.rs
eugenebokhan: done at 9acdf063

crates/backend-uzu/src/backends/metal/kernel/matmul/gemv/gemv.metal
eugenebokhan: done

crates/backend-uzu/src/backends/metal/kernel/matmul/gemv/kernel.rs
eugenebokhan: doublechecked perf, decided to use 8 everywhere

uuuvn (approved):

PR #459, opened 2026-05-28
Implement per layer embedding

CC-Yeh:


CC-Yeh:
The PLE table is kept fully resident (like MLX/vLLM/SGLang) rather than sparse-buffer-backed, since paging a read-only, scatter-accessed weight table would need a demand-residency/streaming layer uzu doesn't have yet (its sparse buffers target the KV-cache append pattern).

crates/backend-uzu/src/backends/metal/kernel/gated_act_mul/gated_act_mul.metal
```diff
@@ -0,0 +1,46 @@
...
+
+template <typename T>
+VARIANTS(T, float, half, bfloat)
+PUBLIC KERNEL(GatedActMul) (
+    const device T* act_operand,
+    const device T* value_operand OPTIONAL(!interleaved),
+    device T* output,
+    const device int32_t* hadamard_factors OPTIONAL(use_hadamard),
+    const constant int& inner_dim,
+    const constant int& outer_dim,
```
uuuvn: Is this hidden_dim batch_dim?

crates/backend-uzu/src/backends/metal/kernel/gated_act_mul/gated_act_mul.metal
```diff
@@ -0,0 +1,46 @@
...
+    const constant int& value_offset,
+    const constant int& value_row_stride,
```
uuuvn: Signed? What's the meaning of those arguments < 0?

crates/backend-uzu/src/backends/metal/kernel/gated_act_mul/gated_act_mul.metal
```diff
@@ -0,0 +1,46 @@
...
+    const constant ActivationType& act_type,
+    const bool interleaved SPECIALIZE,
+    const bool use_hadamard SPECIALIZE,
+    uint col AXIS(inner_dim, 64),
+    uint row AXIS(outer_dim, 1)
```
uuuvn: col/row is hard to keep in mind what they are, I usually do whatever the _dim is but _idx

crates/backend-uzu/src/backends/metal/kernel/gated_act_mul/gated_act_mul.metal
```diff
@@ -0,0 +1,46 @@
...
+  T value;
+  if (interleaved) {
+    int base = row * (2 * inner_dim);
+    value = act_operand[base + col];
+    activated = activate(act_operand[base + inner_dim + col], act_type);
+  } else {
+    activated = activate(act_operand[row * inner_dim + col], act_type);
+    value = value_operand[row * value_row_stride + value_offset + col];
+  }
+  T result = T(float(value) * float(activated));
```
uuuvn: What's the purpose of casts here? I don't think T(float(X) * float(Y)) where X and Y are T does anything at all for any T that is fully representable by float (f16/bf16 are). If the intention is higher precision activation then it should be casted to float before activate, muld, then downcasted.

crates/backend-uzu/src/encodable_block/decoder.rs
```diff
@@ -180,10 +192,51 @@ impl<B: Backend> Decoder<B> {
...
+        inner_features: &Allocation<B>,
+        batch_dim: usize,
+        encoder: &mut Encoder<B>,
+    ) -> Result<Option<Allocation<B>>, DecoderError<B>> {
+        match &self.per_layer_embedding {
+            Some(ple) => {
+                ple.encode(token_ids, inner_features, batch_dim, encoder).map(Some).map_err(DecoderError::BackendError)
+            },
+            None => Ok(None),
+        }
```
uuuvn: if let?

uuuvn (changes requested):
Found a few nits so far, will do a full review tomorrow

crates/backend-uzu/src/backends/metal/kernel/gated_act_mul/gated_act_mul.metal
CC-Yeh: removed casts

crates/backend-uzu/src/backends/metal/kernel/gated_act_mul/gated_act_mul.metal
CC-Yeh: fixed, all `uint` now

crates/backend-uzu/src/backends/metal/kernel/gated_act_mul/gated_act_mul.metal
CC-Yeh: changed to

`outer_dim` -> `batch_dim`
`inner_dim` -> `gated_dim` (for both `hidden_dim` and `ple_dim`)

crates/backend-uzu/src/backends/metal/kernel/gated_act_mul/gated_act_mul.metal
CC-Yeh: fixed, all with `_idx` suffixes now

crates/backend-uzu/src/encodable_block/decoder.rs
CC-Yeh: using `if let` still have similar LoC, used `transpose()` to have less LoC

uuuvn: The goal is readability and maintainability, not LoC. LoC is just a cheap (literally count `\n`s) but very imprecise proxy for it.

Compare:

```rust
match &self.per_layer_embedding {
    Some(ple) => {
        ple.encode(token_ids, inner_features, batch_dim, encoder).map(Some).map_err(DecoderError::BackendError)
    },
    None => Ok(None),
}
```

```rust
self.per_layer_embedding
    .as_ref()
    .map(|ple| ple.encode(token_ids, inner_features, batch_dim, encoder).map_err(DecoderError::BackendError))
    .transpose()
```

```rust
if let Some(ple) = &self.per_layer_embedding {
    ple.encode(token_ids, inner_features, batch_dim, encoder).map(Some).map_err(DecoderError::BackendError)?;
}
```

To me the last one is obviously the easiest to read and maintain one, then match which is just more verbose and more braincycles to parse what's going on, the middle thing is dead last because usually we use this functional-ish style when we care about the result and it doesn't have side effects, here it's kind of abused and I have to spend even more braincycles to understand what's going on.

CC-Yeh:
@uuuvn can you take a look again?

uuuvn (approved):

PR #479, opened 2026-05-31
bench: configurable speculator and acceptance metrics

sady4850:
The benchmark runner already drives `ChatSession`, but always with the default (no-op) speculator, so the speculator metrics that `output.stats` already carries (`speculator_proposed`/`speculator_accepted`, #272) are never exercised or reported.

This adds an optional typed `speculator` field to the benchmark task (`prompt_lookup` for now) so speculation can be measured end-to-end, and surfaces **Accepted / forward pass** (the #272 metric) and **Acceptance rate** in the summary table when a speculator is configured.

The field is `#[serde(default)]` → `None`, so existing task files parse and run unchanged — and the two extra rows only appear when a speculator is set, so non-speculative runs are not cluttered.

### Usage
```json
"speculator": { "kind": "prompt_lookup", "max_ngram_size": 3, "number_of_speculated_tokens": 15 }
```

### Output
Same copy-style prompt (Llama-3.2-1B, greedy), with vs without the field:

**Without `speculator` (unchanged):**
```
╭─────────────────┬───────────────────╮
│ Metric          ┆             Value │
╞═════════════════╪═══════════════════╡
│ Used memory, GB ┆     2.557 ± 0.000 │
│ TTFT, s         ┆     0.078 ± 0.001 │
│ Prompt, t/s     ┆ 1380.514 ± 17.711 │
│ Generate, t/s   ┆    69.946 ± 0.519 │
╰─────────────────┴───────────────────╯
```

**With `prompt_lookup` (n=3, k=15):**
```
╭─────────────────────────┬──────────────────╮
│ Metric                  ┆            Value │
╞═════════════════════════╪══════════════════╡
│ Used memory, GB         ┆    2.557 ± 0.000 │
│ TTFT, s                 ┆    0.084 ± 0.000 │
│ Prompt, t/s             ┆ 1279.186 ± 1.594 │
│ Generate, t/s           ┆  282.662 ± 0.674 │
│ Accepted / forward pass ┆    9.167 ± 0.000 │
│ Acceptance rate, %      ┆   91.667 ± 0.000 │
╰─────────────────────────┴──────────────────╯
```

### DIFF
--- a/crates/benchmarks/src/runner/runner.rs
+++ b/crates/benchmarks/src/runner/runner.rs
@@ -1,20 +1,22 @@
 use std::{
     path::PathBuf,
+    sync::Arc,
     time::{SystemTime, UNIX_EPOCH},
 };
 
 use backend_uzu::{
     VERSION,
+    prelude::PromptLookupSpeculator,
     session::{
         ChatSession,
-        config::{DecodingConfig, RunConfig},
+        config::{DecodingConfig, RunConfig, SpeculatorConfig},
         parameter::{PrefillStepSize, SamplingMethod, SamplingPolicy},
         types::{Input, Output},
     },
 };
 use sysinfo::System;
 
-use crate::runner::types::{Device, Result as TaskResult, Task};
+use crate::runner::types::{Device, Result as TaskResult, SpeculatorSpec, Task};
 
 pub struct Runner {
     pub task: Task,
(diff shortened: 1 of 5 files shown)


uuuvn:
I don't think we want this, currently speculators in uzu are more of a PoC, but soon they won't be and we'll have some other system for managing them (likely by embedding into models)

sady4850:
Thanks for the feedback and the context. When you say "embedding into models" — are you heading towards Medusa/EAGLE-style trained heads (or something like Apple's ReDrafter)? I was thinking of trying a few known speculators here, but it probably makes more sense not to touch that area before you settle on the new system?

PR #487, opened 2026-06-02
Decoding loop benches

agolokoz:


crates/backend-uzu/tests/common/metrics.rs
```diff
@@ -0,0 +1,27 @@
...
+
+pub fn wait_gpu_cooldown() {
+    let mut gpu_temp = get_gpu_temp();
+    if gpu_temp <= COOL_GPU_TEMP {
+        return;
+    }
+
+    println!("Waiting GPU for cooldown...");
+    while gpu_temp > COOL_GPU_TEMP {
+        std::thread::sleep(std::time::Duration::from_millis(10));
```
eugenebokhan: let's avoid very::long::types and import them on top of the file

crates/backend-uzu/tests/performance/session/chat_session_bench.rs
```diff
@@ -0,0 +1,55 @@
...
+    let decoding_config = DecodingConfig::default();
+    ChatSession::new_with_backend::<B>(model_path, decoding_config).unwrap()
+}
+
+fn run_session(
+    session: &mut ChatSession,
+    input: Input,
+) -> Duration {
+    let run_config = RunConfig {
+        tokens_limit: 64,
```
eugenebokhan: shouldn't it be configurable?

crates/backend-uzu/tests/performance/session/chat_session_bench.rs
```diff
@@ -0,0 +1,55 @@
...
+    Duration::from_secs_f64(result.stats.total_stats.duration)
+}
+
+#[uzu_bench]
+fn bench_chat_session(c: &mut Criterion) {
+    let message = "How are you?";
+    for_each_non_cpu_backend!(|B| {
+        wait_gpu_cooldown();
+
+        let mut session = create_session::<B>();
```
eugenebokhan, line 42: Should we create session from scratch each time?

crates/backend-uzu/tests/common/metrics.rs
eugenebokhan: I would rather use tokio that yields control without blocking the thread

```rust
use tokio::time::{sleep, Duration};

async fn await_gpu_cooldown(millis: u64) {
    sleep(Duration::from_millis(millis)).await;
}
```

crates/backend-uzu/tests/performance/session/chat_session_bench.rs
agolokoz: For each backend of course

crates/backend-uzu/tests/common/metrics.rs
agolokoz: Since all callers are non-async, they need to wrap it to sync code and block thread anyway. Or I don't understand anything?

eugenebokhan (approved):

PR #503, opened 2026-06-05
Misc config fixes

kolayne:
- Cargo.toml: the `tokenizers` feature of xgrammar is sufficient (and a bunch of unused dependencies are gone, see Cargo.lock);

- .helix/languages.toml: Fix config parsing errors on helix 25.07.1

### DIFF
--- a/.helix/languages.toml
+++ b/.helix/languages.toml
@@ -1,13 +1,16 @@
 [[language]]
 name = "c"
+scope = "scope.c"
 file-types = []
 
 [[language]]
 name = "cpp"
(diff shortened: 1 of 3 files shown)


uuuvn:
Hm, that's very strange, I have helix 25.07.1 as well and no errors.

uuuvn (approved):

uuuvn:
I had metal language defined in my user-level helix config and uzu project-level config just overwrote .h to be metal too. I've upstreamed the full thing in #504, might need to copy runtime to ~/.config/helix though.

PR #508, opened 2026-06-06
fix benchmark actions

mullakhmetov:
- switch to new benchmark flow trigger params
- disable benchmarks upload by default for a while
- add m3-pax & m4-max pools


uuuvn:
I don't want to have a top-level tests/ that is just benchmark action tests in uzu repo. Can you move it into scripts/ or somewhere else?

mullakhmetov:
@uuuvn I threw them away

uuuvn (approved):

PR #522, opened 2026-06-08
Support  benchmark gpu capture

CC-Yeh:


uuuvn (approved):

crates/backend-uzu/tests/common/mod.rs
```diff
@@ -15,6 +15,18 @@ pub fn type_short_name<T>() -> &'static str {
     std::any::type_name::<T>().rsplit("::").next().unwrap()
 }
 
+pub fn env_var_enabled(name: &str) -> bool {
+    std::env::var(name).is_ok_and(|v| v == "1" || v.eq_ignore_ascii_case("yes") || v.eq_ignore_ascii_case("true"))
+}
+
+pub fn enable_benchmark_gpu_capture_if_requested() {
+    if env_var_enabled("UZU_CAPTURE_BENCH") {
```
eugenebokhan, line 23: let's move all env vars declarations to a dedicated place to avoid strings, wdyt?

CC-Yeh: Yes: https://github.com/trymirai/uzu/pull/524

PR #531, opened 2026-06-11
Add reasoning_effort support to chat completions server

ukint-vs:
## Summary

This adds `reasoning_effort` support to the OpenAI-compatible `/v1/chat/completions` server endpoint. The main goal is to make OpenAI-style clients able to control reasoning/thinking behavior when serving local uzu models such as the Qwen3.5 family.

## Behavior

- Accepts request-level `reasoning_effort` without letting malformed values fail at Rocket JSON extraction time.
- Maps `none` and the existing uzu alias `disabled` to `ReasoningEffort::Disabled`.
- Accepts `default`, `low`, `medium`, and `high` and stores the requested effort on the latest chat message, matching the existing `ChatMessageList::reasoning_effort()` behavior.
- Returns OpenAI-shaped HTTP 400 errors for invalid values.
- Treats recognized OpenAI values that uzu cannot represent today, `minimal` and `xhigh`, as unsupported instead of silently approximating them.

For current local thinking-capable templates, this primarily behaves as a thinking toggle: `none` disables thinking, while non-disabled efforts enable it. The original effort value is still preserved in the chat message so backends/templates that can honor levels can do so.

## Tests

- `cargo test -p cli chat_completions --no-default-features`
- `cargo test -p cli chat_completions --features capability-grammar`
- `git diff --check`


uuuvn:
#695

PR #543, opened 2026-06-15
Optimize chat session stats collection

kolsha:
Hi folks,

I made a small performance optimization in the chat session path. 
Based on my local [benchmarks](https://github.com/kolsha/uzu/tree/kolsha-so-benchmarks) with no-thinking mode, this is around ~7% faster in the relevant structured output scenarios.

Would appreciate a review.


kolsha:
@codex review

eugenebokhan (approved):

PR #545, opened 2026-06-15
Big refactor

uuuvn:
Contains:
- A rewrite of host mixers code, states and speculative suffix accept.
Now it's *significantly* simpler than the distributed state we had
previously with cache layers. This removed the last ugly remnants of
`mod forward_pass;` (cache layers, model shape, token input). Also small
rewrite of rope and attention prepare because the old stuff was a bit
too bad to just copy-paste into the new attention.
- Unification of decoder layer executables and classifier layer. Now
everything common between the two is Transformer and decoder/classifier
are lightweight wrappers. This includes unification of rmsnorm and what
was previously a separate classifier-only "layernorm" (just rmsnorm with
subtract_mean, now a specialization constant in a single normalization
kernel)
- A rewrite of everything down from encodable blocks: language model
generator / classifier, language model generator / classifier context,
sessions and surrounding helpers (like 30 relatively small but very
ugly files). The new replacement lives in engine/ and works on token
level with new message processing crates handling the rest.
- Token stream support in shoji/nagare/hanashi and bridge between new
engine and backend traits (from #565, #571, #575 by @agolokoz)
- Removal of a bunch of dead code that was found while doing all of this

Temporary removals:
- Tracing was removed. It was not working due to lalamo changes, will be
replaced by a new tracing infra.
- TTS was removed. It was a huge unmaintainable vibe coded mess that was
interfering with the refactor. We will likely write a clean qwen3-tts
focused implementation in the future.
- Speculation in prefill is no longer done by the new decoding loop,
will likely be brought back together with contiguous batching

Followups:
- Api misuse hardening. A lot of the new code does not attempt to defend
against invalid inputs. Most of it still shouldn't but some should.
- Consistent u32 token ids in lookup and readout to get rid of token
copy kernel


crates/backend-uzu/src/bridge/chat_token_state.rs
```diff
@@ -0,0 +1,25 @@
...
+            value: SyncShared::new(value),
+        }
+    }
+}
+
+impl<B: Backend> LlmInstanceState for UzuChatTokenBackendInstanceState<B> {
+    fn clone_boxed(&self) -> Box<dyn LlmInstanceState> {
+        Box::new(Self {
+            value: self.value.clone(),
+        })
```
CC-Yeh: nits: Is this used? Can we delete it?
This does not seem like an independent state copy.

crates/backend-uzu/src/engine/language_model/stream/stream.rs
```diff
@@ -0,0 +1,581 @@
...
+
+impl<'a, B: Backend> Iterator for LanguageModelStream<'a, B> {
+    type Item = Result<u64, LanguageModelStreamError<B>>;
+
+    fn next(&mut self) -> Option<Result<u64, LanguageModelStreamError<B>>> {
+        self.generate().transpose()
+    }
+}
+
+impl<'a, B: Backend> Drop for LanguageModelStream<'a, B> {
```
CC-Yeh, line 532: nit: Can we make early drop/cancel invalidate `LanguageModelState` instead of trying to preserve continuation state here?

CC-Yeh: cancelled/dropped streams should force the next request to reset + full-prefill

CC-Yeh (commented):
LGTM 🚀

crates/backend-uzu/src/bridge/chat_token_state.rs
uuuvn: @agolokoz

crates/backend-uzu/src/engine/language_model/stream/stream.rs
uuuvn: Why? This is very much intentional, you create an empty state, create a stream (input = prefill), eat tokens from iterator until you have the response, drop the stream, then the user types a new message, create stream (input = prefill), eat from an iterator until you have the response, etc. LanguageModelState should only be invalidated if there is an error which I haven't implemented yet but will in the follow up hardening pr (a lot of things are very sketchy around post-error states here and in many other places in uzu)

CC-Yeh: Ah yeah, that makes sense. I meant more for error/cancel paths, so maybe the contract is just normal drop commits state, error/cancel invalidates or resets in the hardening PR?

crates/backend-uzu/src/backends/metal/kernel/generated/activation_type.h
```diff
@@ -9,8 +9,7 @@ enum class ActivationType : uint32_t {
   SILU = 0,
   GELUApprox = 1,
   GELUExact = 2,
-  TANH = 3,
```
CC-Yeh, line 12: why remove `TANH`? not used?

crates/backend-uzu/src/backends/metal/kernel/attention/attention_prepare.metal
```diff
@@ -0,0 +1,73 @@
+#include "../common/dsl.h"
+
+// TODO: very ugly and wasteful, clean up and optimize
+
+template <typename ElementT, typename RopeT>
+inline ElementT apply_rope(
```
CC-Yeh: `METAL_FUNC`?

crates/backend-uzu/src/backends/metal/kernel/sampling/context_ring_update.metal
```diff
@@ -0,0 +1,37 @@
...
+
+  device uint32_t* ring_tokens = context_ring + 2;
+  const uint32_t write_start = input_length > suffix_repetition_length ? input_length - suffix_repetition_length : 0;
+
+  for (uint32_t input_index = write_start + thread_index; input_index < input_length; input_index += THREADGROUP_SIZE) {
+    const uint32_t slot = (ring.ring_offset + ring.ring_length + input_index) % suffix_repetition_length;
+    ring_tokens[slot] = static_cast<uint32_t>(input[input_index]);
+  }
+
+  threadgroup_barrier(mem_flags::mem_none);
```
CC-Yeh, line 26: Is this neccessary?

crates/backend-uzu/src/backends/metal/kernel/sampling/context_ring_update.metal
```diff
@@ -0,0 +1,37 @@
...
+    const constant uint32_t& input_length,
+    uint32_t thread_index THREADS(THREADGROUP_SIZE)
+) {
+  const RingParams ring = *reinterpret_cast<device RingParams*>(context_ring);
+
+  device uint32_t* ring_tokens = context_ring + 2;
+  const uint32_t write_start = input_length > suffix_repetition_length ? input_length - suffix_repetition_length : 0;
+
+  for (uint32_t input_index = write_start + thread_index; input_index < input_length; input_index += THREADGROUP_SIZE) {
+    const uint32_t slot = (ring.ring_offset + ring.ring_length + input_index) % suffix_repetition_length;
```
CC-Yeh, line 22: Can `suffix_repetition_length` be zero? if so we should guard it

crates/hanashi/src/classification/encoding.rs
```diff
@@ -0,0 +1,118 @@
...
+            .add_template(TEMPLATE_NAME, config.prompt_template.as_str())
+            .map_err(|_| Error::UnableToEncodeText)?;
+
+        let messages: Vec<HashMap<String, String>> = input
+            .iter()
+            .map(|message| {
+                let role = match message.role {
+                    ClassificationRole::Assistant => config.assistant_role_name.as_str(),
+                    ClassificationRole::User => config.user_role_name.as_str(),
+                };
```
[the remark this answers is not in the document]
agolokoz: There is no such models with `token_codec_config.default_system_prompt`

crates/backend-uzu/src/backends/metal/kernel/attention/attention_prepare.metal
```diff
@@ -0,0 +1,73 @@
...
+    const uint32_t k_idx =
+        (kv_token_offset + batch_idx) * num_kv_heads * head_dim + (head_idx - num_q_heads) * head_dim + head_dim_idx;
+
+    keys[k_idx] = element;
+  } else if (has_kv) {
+    const uint32_t v_idx = (kv_token_offset + batch_idx) * num_kv_heads * head_dim +
+                           (head_idx - num_q_heads - num_kv_heads) * head_dim + head_dim_idx;
+
+    values[v_idx] = element;
+  }
```
eugenebokhan, line 72:
```suggestion
  if (is_query) {
    const uint32_t q_head_idx = head_idx % num_q_heads;

    const uint32_t q_idx =
        q_head_idx * batch_dim * head_dim +
        batch_idx * head_dim +
        head_dim_idx;

    queries[q_idx] = element;
  } else if (has_kv) {
    const uint32_t kv_head_idx = (head_idx - num_q_heads) % num_kv_heads;

    const uint32_t kv_idx =
        (kv_token_offset + batch_idx) * num_kv_heads * head_dim +
        kv_head_idx * head_dim +
        head_dim_idx;

    if (is_key) {
      keys[kv_idx] = element;
    } else {
      values[kv_idx] = element;
    }
  }
```

eugenebokhan (approved):

crates/backend-uzu/src/backends/metal/kernel/generated/activation_type.h
uuuvn: Yes, not used

crates/backend-uzu/src/backends/metal/kernel/sampling/context_ring_update.metal
uuuvn: Yes, otherwise first simd (threads 0..32) can execute the whole thing including the `if (thread_index == 0)` ring metadata update before other simds (threads 32..256) load the metadata `const RingParams ring = *reinterpret_cast<device RingParams*>(context_ring);` and it will screw everything up

crates/backend-uzu/src/backends/metal/kernel/sampling/context_ring_update.metal
uuuvn: It shouldn't be zero, suffix_repetition_length = None is the right representation of no suffix rep

crates/backend-uzu/src/backends/metal/kernel/attention/attention_prepare.metal
uuuvn: `const uint32_t q_head_idx = head_idx % num_q_heads;` is dead code in the query branch because query is `head_idx < num_q_heads`. The kv part looks a bit cleaner, will prob do it like this when I rewrite this kernel to be more efficient (currently it does the same suboptimal read twice as our old rope kernel)

crates/backend-uzu/benches/session/chat_session_bench.rs
```diff
@@ -1,58 +0,0 @@
-use std::time::Duration;
```
uuuvn, line 1: This should be brought back somehow. It doesn't fit in uzu, but backend-uzu can't render/tokenize. I think the whole architecture of message processing crates is wrong, there should be one message-level trait and then a message processing wrapper that backend-uzu wraps itself in

crates/backend-uzu/src/array/element.rs
```diff
@@ -2,16 +2,21 @@ use bytemuck::Pod;
 use half::{bf16, f16};
 use num_traits::NumCast;
 
+#[cfg(test)]
 use crate::data_type::DataType;
 
+// TODO: remove this
```
uuuvn, line 8: Shouldn't forget about this. There is no longer an array, ArrayElement trait should be removed too

crates/backend-uzu/src/backends/common/kernel/mod.rs
```diff
@@ -1,7 +1,58 @@
-use crate::backends::common::Backend;
+use crate::backends::common::{AsBufferRangeMut, AsBufferRangeRef, Backend, Buffer};
 
 pub mod matmul;
 
+// TODO: This whole thing is horrible and should be unified with AsBufferRangeRef
```
uuuvn, line 5: Unified BufferRange?

crates/backend-uzu/src/backends/common/encoder.rs
```diff
@@ -33,14 +34,20 @@ fn resolve_copy_range(
 pub struct Encoder<'encoding, B: Backend> {
     context: &'encoding B::Context,
     command_buffer: <B::CommandBuffer as CommandBuffer>::Encoding,
-    allocation_pool: AllocationPool<B>,
+    allocation_pool: Rc<AllocationPool<B>>,
```
uuuvn, line 37: Should be asref? Definitely shouldn't be rc

uuuvn: Pools are the wrong abstraction I think, there should be better tracking of resource used in command buffer. Idk what it's going to look like though

crates/backend-uzu/src/config/model/dflash_speculator.rs
```diff
@@ -2,6 +2,7 @@ use proc_macros::uzu_config;
 
 use crate::config::dflash::DFlashDraftConfig;
 
+#[allow(unused)] // TODO
```
uuuvn, line 5: Should be removed when dflash is implemented

crates/backend-uzu/src/encodable_block/mixer/attention/core/fallback.rs
```diff
@@ -0,0 +1,174 @@
...
+    softmax: <B::Kernels as Kernels>::SoftmaxKernel,
+    matmul: RefCell<<B::Kernels as Kernels>::MatmulKernel>,
+}
+
+impl<B: Backend> AttentionFallbackCore<B> {
+    pub fn new(
+        arguments: &AttentionCoreNewArguments,
+        context: &B::Context,
+    ) -> Result<Self, B::Error> {
+        assert!(!arguments.is_trie, "trie not supported by attention fallback"); // Is it?
```
uuuvn, line 34: Is it?

crates/backend-uzu/src/encodable_block/mixer/attention/rope.rs
```diff
@@ -0,0 +1,103 @@
...
+    pub sines: Allocation<B>,
+    pub dim: usize,
+}
+
+impl<B: Backend> PrecalculatedRoPE<B> {
+    pub fn precalculate(
+        rope_config: &AnyRoPEConfig,
+        token_positions: &[usize],
+        encoder: &mut Encoder<B>,
+    ) -> Result<Self, B::Error> {
```
uuuvn, line 17: This being gpu is required for fully gpu speculation

crates/backend-uzu/src/encodable_block/mixer/attention/state.rs
```diff
@@ -0,0 +1,236 @@
...
+        };
+
+        Some(RingParams {
+            ring_offset: *offset as u32,
+            ring_length: *length as u32,
+        })
+    }
+}
+
+pub struct AttentionState<B: Backend> {
```
uuuvn, line 54: Should be much more defensive against misuse

crates/backend-uzu/src/encodable_block/mixer/mod.rs
```diff
@@ -0,0 +1,126 @@
...
+        max_context_length: Option<usize>,
+        context: &B::Context,
+    ) -> Result<Box<dyn MixerState<B>>, B::Error>;
+
+    fn encode(
+        &self,
+        hidden: Allocation<B>,
+        precalculated_rope: Option<&PrecalculatedRoPE<B>>,
+        batch_dim: &BatchTopology,
+        state: Option<MaybeMut<dyn MixerState<B>>>,
```
uuuvn, line 67: All mixers should have some mechanism to check that mixer state was created by them

crates/backend-uzu/src/encodable_block/embedding.rs
```diff
@@ -114,14 +113,16 @@ struct LogitSoftCap<B: Backend> {
     kernel: <B::Kernels as Kernels>::LogitSoftCapKernel,
 }
 
+// TODO: embedding lookup dtype (u64) should match sampling (u32)
```
uuuvn, line 116: embedding lookup dtype (u64) should match sampling (u32)

crates/backend-uzu/src/encodable_block/transformer.rs
```diff
@@ -0,0 +1,243 @@
...
+    }
+
+    pub fn max_context_length(&self) -> Option<usize> {
+        self.layers.iter().map(|(layer, _rope_index)| layer.mixer.max_context_length()).fold(None, |acc, el| {
+            match (acc, el) {
+                (Some(a), Some(b)) => Some(usize::min(a, b)),
+                (Some(x), None) | (None, Some(x)) => Some(x),
+                (None, None) => None,
+            }
+        })
```
uuuvn, line 155: should be shared and simpler

crates/backend-uzu/src/encodable_block/transformer_layer.rs
```diff
@@ -0,0 +1,233 @@
...
+    Normalization(#[from] NormalizationNewError<B>),
+    #[error("Layer {layer_index} sets post_layer_scalar but has no post_mlp_norm")]
+    PostLayerScalarWithoutPostMlpNorm {
+        layer_index: usize,
+    },
+    #[error("Transformer layers except the first one if it doesn't have rht in mixer require pre_mixer_norm_config")]
+    MissingPreMixerNormConfig,
+}
+
+// TODO: saner shortcut
```
uuuvn, line 36: saner shortcut

crates/backend-uzu/src/encodable_block/transformer_layer.rs
```diff
@@ -0,0 +1,233 @@
...
+    ) -> Result<Allocation<B>, B::Error> {
+        let hidden = if let Some(pre_mixer_norm) = &self.pre_mixer_norm {
+            pre_mixer_norm.encode(&input, 0, batch_dim.size(), Some(shortcut), encoder)?
+        } else {
+            assert!(self.layer_index == 0);
+            encoder.encode_copy(&input, .., shortcut, ..);
+            input
+        };
+
+        // TODO: In prefill outside of sampling suffix in last layer part of mixer (ie out projection) and everything after is dead code
```
uuuvn, line 210: In prefill outside of sampling suffix in last layer part of mixer (ie out projection) and everything after is dead code

CC-Yeh (approved):

agolokoz (approved):

PR #546, opened 2026-06-16
Fix offline Mirai model cache resolution

kolsha:
I wanted to test inference while sitting on the airplane, but helaas pindakaas, it could not resolve the model without internet.

---
- Added caching capabilities for model metadata in the Mirai registry.
- Implemented snapshot loading and saving to persist registry state.
- Updated model fetching logic to utilize cached data when available.
- Introduced tests to ensure caching behavior does not store errors.
- Refactored storage configuration to support new cache paths and model metadata persistence.

@codex review


eugenebokhan:
The latest main contains register caching

PR #553, opened 2026-06-18
Fix HadamardTransform Slang kernel signature

KonstantinSKY:
Adds the missing `transform_order` specialize parameter to the Slang `HadamardTransform` so its signature matches CPU/Metal. Without it, the build-script's cross-backend signature check fails under `--features webgpu`.


uuuvn:
This branch is internal/experimental, it doesn't work yet. Also the kernel stubs are autogenerated via a python script, I don't remember if I committed it or not, but when we get to rebasing and reviving this branch we'll just re-run the script because there was *a lot* of other kernel changes in main kernels that should be just re-autogened instead of fixing each one by hand

KonstantinSKY:
Thanks for the context, makes sense.

PR #562, opened 2026-06-24
Power Sampling Stats

eugenebokhan:


crates/backend-uzu/src/session/types/stats.rs
```diff
@@ -1,5 +1,86 @@
...
+                        total += watts;
+                        watts
+                    },
+                    None => continue,
+                },
+            };
+            samples_count += 1;
+            package += wall_watts;
+            max_package = max_package.max(wall_watts);
+            energy_joules += wall_watts * window_secs;
```
CC-Yeh, line 63: This feels wrong. `window_secs` is gap between sample-completion timestamps.

crates/backend-uzu/src/session/types/stats.rs
```diff
@@ -1,5 +1,86 @@
...
+            samples_count,
+            average_cpu_watts: cpu / sample_count,
+            average_gpu_watts: gpu / sample_count,
+            average_gpu_sram_watts: gpu_sram / sample_count,
+            average_ane_watts: ane / sample_count,
+            average_ram_watts: ram / sample_count,
+            average_total_watts: total / sample_count,
+            average_package_watts: package / sample_count,
+            max_package_watts: max_package,
+            energy_joules,
```
CC-Yeh, line 81: `avg_watts` x duration != energy_joules since one is sample averaged. Is this gap intentional?

crates/keisoku/src/snapshot.rs
```diff
@@ -23,4 +23,45 @@ pub struct Snapshot {
...
     pub sensors: Vec<Sensor>,
+    /// HID voltage rails; on iOS the only un-gated electrical signal, empty elsewhere.
+    #[serde(default)]
+    pub voltage: Vec<Sensor>,
+    /// HID current rails, paired with [`Snapshot::voltage`] by rail code.
+    #[serde(default)]
+    pub current: Vec<Sensor>,
+}
+
+impl Snapshot {
```
CC-Yeh, line 34: nit: This could grow huge, some cap or ring buffer?

crates/keisoku/src/recorder/handle.rs
```diff
@@ -3,87 +3,63 @@ use std::{
...
-    let markers = Arc::new(Mutex::new(Vec::new()));
+    let snapshots = Arc::new(Mutex::new(Vec::new()));
 
-    let worker = {
+    {
         let stop_flag = Arc::clone(&stop_flag);
-
+        let snapshots = Arc::clone(&snapshots);
         std::thread::spawn(move || {
             let mut collector = Collector::new();
```
CC-Yeh, line 55: Can we move `Collector` construction before `started_at` to avoid first window skewed?
Not sure how expensive is it though. Some IO setups

crates/keisoku/src/recorder/handle.rs
```diff
@@ -3,87 +3,63 @@ use std::{
...
         }
     }
 }
 
 impl Drop for RecorderHandle {
     fn drop(&mut self) {
         self.stop_flag.store(true, Ordering::Relaxed);
-        if let Some(worker) = self.worker.take() {
-            let _ = worker.join();
-        }
```
CC-Yeh, line 56: why this is deleted? I think this would allow clean exit?

crates/backend-uzu/src/session/types/stats.rs
eugenebokhan: Yes, that gap is the real time each reading covers, so `watts × gap` gives energy

crates/backend-uzu/src/session/types/stats.rs
eugenebokhan: Good catch, that was the bug. Averages are now time-weighted, so `avg × total_time = energy` exactly. Fixed.

crates/keisoku/src/snapshot.rs
eugenebokhan: It's per-response only — ~600 entries for a 60s reply, then cleared. Stays small, but we can add a cap if you want one

crates/keisoku/src/recorder/handle.rs
eugenebokhan: Put it back

crates/keisoku/src/recorder/handle.rs
eugenebokhan: Done

CC-Yeh (approved):
LGTM

Screenshot 2026-06-29 at 9 07 19 PM

PR #565, opened 2026-06-25
Add token streaming

agolokoz:


crates/backend-uzu/src/bridge/chat_token_backend.rs
```diff
@@ -0,0 +1,213 @@
...
+
+    fn stop_token_ids(&self) -> Option<Box<[u64]>> {
+        Some(self.stop_token_ids.iter().map(|id| *id as u64).collect())
+    }
+}
+
+/// Wraps a stream to assert it is `Send`.
+///
+/// The generated decoding stream holds `MutexGuard`s (which are `!Send`) across
+/// yield points, but the underlying model is driven from a single thread and the
```
CC-Yeh: Is this true?
```suggestion
/// but the underlying model is driven from a single thread
```
I see there's `tokio`'s multi-thread runtime, a task might get resumed from different thread.

agolokoz: You are right. Removed this comment

crates/backend-uzu/src/bridge/sync_shared.rs
```diff
@@ -0,0 +1,29 @@
...
+
+impl<T> Clone for SyncShared<T> {
+    fn clone(&self) -> Self {
+        Self {
+            value: self.value.clone(),
+        }
+    }
+}
+
+unsafe impl<T> Send for SyncShared<T> {}
```
CC-Yeh, line 27: is this actually safe? SyncShared is Send/Sync for any T but the inner state is !Send. if we hold the lock or an Rc across an await, the task can move threads and we're in UB territory. or does the guard always drop before the await? couldn't tell from reading it.

uuuvn (approved):

crates/backend-uzu/src/bridge/sync_shared.rs
uuuvn: Almost definitely technically wrong, probably won't blow up in practice, this is a workaround for now just to get something working to merge the refactor, this should be (and will be shortly after) just a carefully threaded Send + Sync bound on all the relevant backend traits

PR #576, opened 2026-07-02
Support `TreeUpdateSolve` kernel

CC-Yeh:
- Rework GDN tree-update solve: fuse kh0 into gram + pack A into f32 block-pair tiles, up to 1.6× faster
- MXU BV32 solve backend via fragment matmul, up to 1.52× faster
- Split gram into dedicated tree_gram + tree_update_solve kernels (Metal + CPU) + dispatch helper/test
- Keep kh0 read-out in f32 (only otherwise-lossy step; matters on long contexts)
- Factor gram/kh0 matmuls into reusable tree_gram_dots / tree_kh0 helpers so a future per-key-head split is localized
- fragment.h helper for MXU
- Simdgroup-scoped device fence in the solve's serial loop
- Remove L2 norm from gram, should fuse with projection instead
- Expand tests (larger trees, realistic A magnitudes) + benches


CC-Yeh:
cc @ry2009

crates/backend-uzu/src/backends/metal/kernel/gdn_tree_verify/tree_gram.metal
```diff
@@ -0,0 +1,287 @@
...
+    threadgroup float row_prefix[ROW_TILE],
+    threadgroup float row_beta[ROW_TILE],
+    threadgroup uint row_token[ROW_TILE],
+    threadgroup float col_prefix[NUM_SIMDGROUPS * COL_TILE],
+    threadgroup uint col_trie_start[NUM_SIMDGROUPS * COL_TILE],
+    threadgroup uint col_trie_end[NUM_SIMDGROUPS * COL_TILE],
+    const bool use_h0 SPECIALIZE,
+    const ThreadContext thread_context,
+    const uint batch_idx GROUPS(batch_size),
+    const uint value_head_idx GROUPS(value_heads),
```
ry2009, line 187: this makes the gram per value head, so the k-gram tile + A storage get duplicated G x (G=2 on qwen3.6 shapes) vs the old per-key-head gram. it's the price of fusing kh0 in -- we hit the same fork on the cuda side and went the other way (per-key-head gram + a separate parallel kh0 kernel) to keep A deduped. did you measure fused-per-hv vs per-key-head + standalone kh0? if not there might be another 10-20% hiding behind the 1.6x

crates/backend-uzu/src/backends/metal/kernel/gdn_tree_verify/tree_update_solve.metal
```diff
@@ -0,0 +1,177 @@
...
+
+using namespace metal;
+using namespace uzu::matmul;
+
+// Block forward substitution for (I + A) U = beta * (v - exp(prefix) * kh0).
+// Per 16-token block i, serially: acc_i = beta * (v_i - exp(prefix_i) * kh0_i);
+// acc_i -= A[i, j] @ U[j] for all j < i; U[i] = Ainv[i] @ acc_i.
+//
+// Shapes (NB = ceil(T/16)):
+//   kh0    [B, T, HV, head_v_dim] in T, from BuildTreeGram (absent when !use_h0)
```
ry2009: kh0 lands in model dtype here (bf16) -- on the cuda side we keep the kh0 workspace f32. it's the only added rounding in the whole pipeline and it scales with the state readout, which grows with context: on long prompts e^G * kh0 can dwarf v, so 0.4% of a big number can exceed the thing you subtract it from. the buffer is tiny -- can we just make it f32? or if bf16 is deliberate for bandwidth, a long-context real-scale error number would settle it (the 1e-2 mxu tolerance would hide a systematic 5e-3 shift)

crates/backend-uzu/src/backends/metal/kernel/gdn_tree_verify/tree_update_solve.metal
```diff
@@ -0,0 +1,177 @@
...
+//          (block (i,j) at pair j/2, column (j%2)*16)
+//   Ainv   [B*HV, NB, 16, 16] f32 diagonal-block inverses from BuildTreeGram
+//   h0_idx [B] i32, negative = zero initial state (absent when !use_h0)
+//   U      [B*HV, T, head_v_dim] f32 (output)
+//
+// Grid: one simdgroup per (batch, value head, BV value-dim tile); the history is
+// consumed one packed pair tile ([16,32] A x [32,BV] U) per fragment_mma plus a
+// single-block tail when i is odd.
+// ponytail: A stays f32; bf16 A only pays in bandwidth-bound large-batch shapes.
+template <typename T, uint BV, bool USE_MXU>
```
ry2009, line 30: ponytail lol. also this comment is load-bearing -- the PR description still says "pack the pair A matrix (bf16)" which contradicts it, pls update the desc so nobody ships bf16 A off a skim

crates/backend-uzu/src/backends/metal/kernel/gdn_tree_verify/tree_update_solve.metal
```diff
@@ -0,0 +1,177 @@
...
+    BlockMatrixFragment inv_frag;
+    TileFragment solved;
+    solved.clear();
+
+    const device float* inv_block = a_inv_blocks + block_idx * (BT * BT);
+    inv_frag.load_from(lane_idx, fragment_source(inv_block, BT).bounded(tile_rows, tile_rows));
+    fragment_mma(solved, inv_frag, acc);
+    solved.store_safe(lane_idx, u_head_tile + row_base * head_v_dim, head_v_dim, short2(tile_value_cols, tile_rows));
+
+    threadgroup_barrier(mem_flags::mem_device);
```
ry2009: threadgroup here is a single simdgroup, so simdgroup_barrier(mem_flags::mem_device) should be enough -- marginally cheaper per block and this loop is the serial path

crates/backend-uzu/unit/backends/common/kernel/tree_update_solve_test.rs
```diff
@@ -0,0 +1,205 @@
...
+
+    let kh0 = (0..v_len).map(|i| cast(((i as f32 * 0.019).sin() * 0.2) + 0.01)).collect::<Vec<_>>();
+    let v = (0..v_len).map(|i| cast(((i as f32 * 0.017).cos() * 0.18) - 0.02)).collect::<Vec<_>>();
+    let prefix = (0..scalar_len)
+        .map(|i| -((i % tree_size) as f32) * 0.01 - ((i % num_v_heads) as f32) * 0.003)
+        .collect::<Vec<_>>();
+    let beta = (0..scalar_len).map(|i| 0.25 + ((i as f32 * 0.013).sin() + 1.0) * 0.2).collect::<Vec<_>>();
+    // Packed [B * HV, NB, ceil(NB/2), BT, 2*BT] block-pair tiles; strictly-lower
+    // blocks get values, everything else zero (the kernel never reads it).
+    let a_f32 = (0..a_len)
```
ry2009, line 122: these synthetic A entries are ~0.01 with the cpu port of the same algorithm as the oracle -- great for catching indexing/layout bugs (it would), but it never stresses conditioning. real A sits at ||A||~1 with l2-normed keys + softplus decays, and the deep (I+A)^-1 cascades are where precision actually dies -- that's what forced tf32x3 on the cuda side. can we add one end-to-end case against a true sequential-recurrence oracle at real magnitudes? i'll send you the exact input recipe from our validation kit (random_tree + normalized keys + softplus decay + h0 ~0.3 scale), ports in ~30 min

crates/backend-uzu/unit/backends/common/kernel/tree_update_solve_test.rs
```diff
@@ -0,0 +1,205 @@
...
+        head_v_dim: 128,
+    },
+    SolveCase {
+        // NB=7: exercises pair_idx >= 1, tail blocks >= 4, and a ragged final block.
+        name: "ragged_seven_blocks",
+        batch_size: 1,
+        tree_size: 97,
+        num_v_heads: 2,
+        head_v_dim: 64,
+    },
```
ry2009, line 69: 97 is a nice case (pairs + tail + ragged all at once) but nothing goes past NB=7. add a T=128 or 256 so num_full_pairs >= 3 gets exercised -- the large-tree regime is the whole reason this kernel exists

ry2009 (commented):
went through the whole thing against the cuda reference (cpu ref, both metal kernels, tests). semantics line up exactly with the validated triton path -- same rhs factorization, same block substitution, ragged guards all trace clean. the pair-packed A + kh0-in-gram is the same optimization we just landed on the cuda side (we hoisted kh0 to its own parallel kernel instead bc our gram is per key head -- same tradeoff, opposite layout, so the two backends now cross-validate each other, which is exactly what the reference relationship is for). left inline asks -- the main ones are a real-scale oracle test, kh0 dtype, and the per-value-head gram duplication question. nice work, the 97-token ragged test case is a good catch.

crates/backend-uzu/unit/backends/common/kernel/tree_update_solve_test.rs
CC-Yeh: @ry2009 where can I find that recipe

crates/backend-uzu/src/backends/metal/kernel/gdn_tree_verify/tree_gram.metal
CC-Yeh: Interesting, will try to bench, now for T=64, B=1 case, all kernels on M5 max are ~10us, not sure adding a new kernel with dispatch overhead etc. will make separation worthy.

Let me try.

crates/backend-uzu/unit/backends/common/kernel/tree_update_solve_test.rs
CC-Yeh: Is it in `sglang-mirai/test/registered/spec/utils/test_gdn_chunk_tree_verify.py`?

crates/backend-uzu/src/backends/metal/kernel/gdn_tree_verify/tree_gram.metal
CC-Yeh: For MXU, the fused one is always better

Screenshot 2026-07-03 at 10 59 35 AM

CC-Yeh: For simdgroup on M5 Max, split wins starting from T >= 128
Screenshot 2026-07-03 at 11 01 19 AM

CC-Yeh: chips sweep
Screenshot 2026-07-03 at 12 11 53 PM

CC-Yeh: We can of course have two implementations (fused and split) but are we really using T >= 128.
If not, I'd prefer keeping fused one, so binary size of uzu stay lean.

crates/backend-uzu/unit/backends/common/kernel/tree_update_solve_test.rs
CC-Yeh: Bumped the synthetic A to ~0.05 so it's off the toy scale.

On the full independent oracle: I ported your recipe and it passes, but I'm going to hold off landing it. It's ~350 LoC for a chain that isn't even wired into the model path yet, and traces will give us the real end-to-end check at integration anyway (against the actual model, not my re-derivation of it). Mostly a maintainability call. Keeping your recipe for when that lands.

One thing I'm curious about though: we solve (I+A)U=b with exact block forward-substitution in f32, not a Neumann series, so does ‖A‖~1 actually bite us the way it forced tf32x3 on your side?

uuuvn (approved):

PR #582, opened 2026-07-06
Generalize attention Q/KV modes

CC-Yeh:
- Makes borrowed-cache query-only attention explicit for Gemma4-style sharing.
  - Adds split Q/KV primitives needed by DFlash:
      - ephemeral suffix attend
      - KV-only append
      - non-causal full KV state


crates/backend-uzu/src/encodable_block/mixer/attention/mode.rs
```diff
@@ -0,0 +1,352 @@
...
+            core::AttentionCoreEncodeArguments,
+            qkv_norm::QKVNorm,
+            rope::PrecalculatedRoPE,
+            state::{AttentionState, AttentionStateType},
+        },
+    },
+    utils::maybe_mut::MaybeMut,
+};
+
+pub(super) type AttentionPrepare<B> = <<B as Backend>::Kernels as Kernels>::AttentionPrepareKernel;
```
uuuvn: This obscures what AttentionPrepare is by making it look like a block where in fact it is a kernel

uuuvn (changes requested):

crates/backend-uzu/src/encodable_block/mixer/attention/mode.rs
CC-Yeh: Replaced it with `PrepareKernel`

uuuvn (approved):

PR #586, opened 2026-07-07
Metal: Add dynamic caching capability check

CC-Yeh:
- Since M3 chips, metal supports dynamic SM memory between register/threadgroup/buffer etc.
- Some cleanups

resource:
https://www.digitaltrends.com/computing/apple-dynamic-caching-explained/
https://developer.apple.com/videos/play/tech-talks/111375/?utm_source=chatgpt.com

### DIFF
--- a/crates/backend-uzu/src/backends/metal/context.rs
+++ b/crates/backend-uzu/src/backends/metal/context.rs
@@ -52,6 +52,12 @@ impl MetalContext {
         self.device.supports_mxu()
     }
 
+    // TODO: remove allow once a dynamic-caching dispatcher uses this probe.
+    #[allow(dead_code)]
+    pub fn supports_dynamic_caching(&self) -> bool {
+        self.device.supports_family(metal::MTLGPUFamily::Apple9)
+    }
+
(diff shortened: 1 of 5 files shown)


CC-Yeh:
AttentionGemm / support_mxu cleanups will be in another branch

crates/backend-uzu/src/backends/metal/context.rs
```diff
@@ -52,6 +52,12 @@ impl MetalContext {
         self.device.supports_mxu()
     }
 
+    // TODO: remove allow once a dynamic-caching dispatcher uses this probe.
+    #[allow(dead_code)]
+    pub fn supports_dynamic_caching(&self) -> bool {
+        self.device.supports_family(metal::MTLGPUFamily::Apple9)
+    }
```
uuuvn, line 59: What's the point of adding this small unused helper? Better add it when it's going to be used, it's 3 lines, there isn't a good reason to upstream it separately.

uuuvn (changes requested):
I see this pr is now refactoring tests and not doing anything related to dynamic caching, can you rename/redescription commit and pr to reflect that?

PR #587, opened 2026-07-07
Fix `support_mxu` leak and refactor `AttentinonGemm` dispatch

CC-Yeh:
- Fix `support_mxu` leak in backend
- Apply GEMM dispatch style to AttentionGemm
- Fix `TreeUpdate`/`TreeGram` dispatch tests


CC-Yeh:
Most Dispatch Code could be eliminated by future `AUTOTUNE` support

crates/backend-uzu/src/backends/common/kernel/attention_gemm/kernel.rs
```diff
@@ -0,0 +1,37 @@
...
+        Backend, BufferArg, BufferArgMut, Encoder,
+        kernel::{Kernels, attention_gemm::AttentionGemmArgs},
+    },
+    data_type::DataType,
+};
+
+pub trait AttentionGemmDispatch: Sized {
+    type Backend: Backend<Kernels: Kernels<AttentionGemmDispatch = Self>>;
+
+    #[allow(clippy::too_many_arguments)]
```
uuuvn: We have this disabled in workspace Cargo.toml

crates/backend-uzu/src/backends/common/kernel/attention_gemm/arguments.rs
```diff
@@ -0,0 +1,15 @@
+use crate::backends::common::gpu_types::{attention::AttnParams, ring::RingParams};
+
+pub struct AttentionGemmArgs<Q, K, V, O, T, S> {
```
uuuvn: We usually add trait bounds to structs too so wrong argument type errors when creating the struct instead of when trying to pass AttentionGemmArgs with unsupported Q for example into the function

uuuvn: Not sure how verbose it's going to be, might make sense to skip if it's like very ugly but otherwise better to specify what the struct is supposed to have in it's decl instead of just in functions it's used in

crates/backend-uzu/src/backends/common/kernel/attention_gemm/kernel.rs
```diff
@@ -0,0 +1,37 @@
...
+        bd: u32,
+        is_kv_cache_ring: bool,
+        is_causal: bool,
+        is_trie: bool,
+        is_sliding_window: bool,
+        has_sinks: bool,
+    ) -> Result<Self, <Self::Backend as Backend>::Error>;
+
+    fn encode<'q, 'k, 'v, 'o, 'trie, 'sinks>(
+        &mut self,
```
uuuvn: Encode should not be &mut self where possible, old attention gemm core didn't have it and this shouldn't too. Afaik this is currently &mut only in matmul and things that wrap it but this is a very bad pattern that we shouldn't copy to other kernels that can avoid it (like if there is a small number of possible variants)

crates/backend-uzu/src/backends/common/kernel/attention_gemm/arguments.rs
uuuvn: Also can this be just AttentionCoreEncodeArguments?

crates/backend-uzu/src/encodable_block/mixer/attention/core/gemm.rs
```diff
@@ -1,84 +1,56 @@
...
         arguments: &AttentionCoreNewArguments,
-        use_mxu: bool,
         context: &B::Context,
     ) -> Result<Self, B::Error> {
-        let bk = if arguments.head_dim < 128 || use_mxu {
+        let bk = if arguments.head_dim < 128 {
             32
         } else {
             16
         };
```
uuuvn: This should be backend-specific. I think the correct layer to separate it at is attention core api. Unless I'm missing something the whole attention core api can be copied mostly as-is

crates/backend-uzu/src/encodable_block/mixer/attention/core/gemm.rs
```diff
@@ -87,49 +59,41 @@ impl<B: Backend> AttentionGemmCore<B> {
...
+                params: AttnParams {
+                    q_strides: [0, (arguments.suffix_length * self.head_dim) as u64, self.head_dim as u64],
+                    k_strides: [0, self.head_dim as u64, (self.num_groups * self.head_dim) as u64],
+                    v_strides: [0, self.head_dim as u64, (self.num_groups * self.head_dim) as u64],
+                    o_strides: [0, self.head_dim as u64, (self.num_q_heads * self.head_dim) as u64],
+                    gqa_factor: (self.num_q_heads / self.num_groups) as u32,
+                    scale: self.scale.unwrap_or(1.0f32 / (self.head_dim as f32).sqrt()),
+                    q_len: arguments.suffix_length as u32,
+                    k_len: (arguments.state_type.physical_prefix_length() + arguments.suffix_length) as u32,
+                    q_off: arguments.state_type.physical_prefix_length() as u32,
```
uuuvn: Like this stuff just screams that it should be packed into backend

crates/backend-uzu/src/backends/common/kernel/attention_gemm/kernel.rs
```diff
@@ -0,0 +1,37 @@
...
+    backends::common::{
+        Backend, BufferArg, BufferArgMut, Encoder,
+        kernel::{Kernels, attention_gemm::AttentionGemmArgs},
+    },
+    data_type::DataType,
+};
+
+pub trait AttentionGemmDispatch: Sized {
+    type Backend: Backend<Kernels: Kernels<AttentionGemmDispatch = Self>>;
+
```
uuuvn, line 13: Let's have `fn is_supported/available(context: &<Self::Backend as Backend>::Context) -> Result<bool, B::Error>` or something like that and delete cpu gemm attention as cpu perf is irrelevant and we can do single pass there.

Ideally maybe the whole attention core system should be in backends with maybe common shared base for similar backends that all have similar dispatching logic, but ideally eg cpu should be just simple single pass and nothing else. This is probably too much for this pr and a future side quest after speculation, just leaving thoughts here

crates/backend-uzu/src/backends/metal/kernel/attention.rs
```diff
@@ -0,0 +1,139 @@
+use std::collections::{HashMap, hash_map::Entry};
+
+use super::AttentionGemmMetalKernel;
```
uuuvn: We avoid super:: imports

uuuvn (changes requested):

crates/backend-uzu/src/backends/common/kernel/attention_gemm/kernel.rs
uuuvn: If it pre-creates a lot then the lazy creation logic should be internal and not leak via &mut self

crates/backend-uzu/src/backends/common/kernel/attention_gemm/kernel.rs
CC-Yeh: removed

crates/backend-uzu/src/backends/common/kernel/attention_gemm/kernel.rs
CC-Yeh: Improved thx

crates/backend-uzu/src/backends/common/kernel/attention_gemm/arguments.rs
CC-Yeh: improved thx

crates/backend-uzu/src/encodable_block/mixer/attention/core/gemm.rs
CC-Yeh: moved

crates/backend-uzu/src/backends/common/kernel/attention_gemm/kernel.rs
```diff
@@ -0,0 +1,49 @@
...
+    ) -> Result<Self, B::Error>;
+
+    fn encode<'a, KT: BufferArg<'a, B>, VT: BufferArg<'a, B>>(
+        &self,
+        arguments: AttentionCoreEncodeArguments<'a, B, KT, VT>,
+        encoder: &mut Encoder<B>,
+    ) -> Result<Allocation<B>, B::Error>;
+}
+
+impl<B> AttentionGemmCore<B> for ()
```
uuuvn: I think something like the (currently unstable) [never type](https://doc.rust-lang.org/std/primitive.never.html) which can't be constructed fits much better as a "kernel not supported" type than () which can be constructed and called encode on (which makes little sense). There is a stable std::convert::Infallible that I would've probably used here but it's a tiny thing, so just leaving as a comment and not a change request

crates/backend-uzu/src/backends/common/kernel/attention_gemm/kernel.rs
```diff
@@ -0,0 +1,49 @@
...
+where
+    B: Backend<Kernels: Kernels<AttentionGemmCore = ()>>,
+{
```
uuuvn: We usually do `<B: Backend<Kernels: Kernels<AttentionGemmCore = ()>>>` to define the generic and it's bound in one place, but it's a small stylistic thing

crates/backend-uzu/src/backends/common/kernel/attention_gemm/kernel.rs
```diff
@@ -0,0 +1,49 @@
+use crate::{
+    backends::common::{Allocation, Backend, BufferArg, Encoder, kernel::Kernels},
+    encodable_block::mixer::attention::core::{AttentionCoreEncodeArguments, AttentionCoreNewArguments},
+};
+
+pub trait AttentionGemmCore<B: Backend<Kernels: Kernels<AttentionGemmCore = Self>>>: Sized {
```
uuuvn, line 8: This trait is also not gemm specific so I think it makes sense to name it just AttentionCore and then `type AttentionGemmCore: attention_core::AttentionCore<Self::Backend>;`

crates/backend-uzu/src/backends/cpu/kernel/attention/attention_gemm.rs
```diff
@@ -32,101 +30,5 @@ pub fn attention_gemm<T: ArrayElement + Float, const BK: u32, const BD: u32, con
...
-            // Normalize and write output
-            let inv_sum = 1.0 / sum_exp;
-            for j in 0..head_dim {
-                unsafe {
-                    *o_row.add(j) = T::from(o_acc[j] * inv_sum).unwrap();
-                }
-            }
-        }
-    }
+    unreachable!("CPU AttentionGemm is only generated to satisfy the shared kernel trait")
```
uuuvn: This metal kernel just shouldn't be a `PUBLIC` instead of a stub in cpu with metal-specific things like `const USE_MXU: bool`

uuuvn (changes requested):
One last real change request, but I've also left a couple comments on small stylistic things I would've done differently, feel free to ignore if you don't agree with something / don't feel like it's worth the change

crates/backend-uzu/src/backends/cpu/kernel/attention/attention_gemm.rs
uuuvn: Poor wording from me here: metal kernel shouldn't be `PUBLIC`, if it isn't public it doesn't add itself into common traits, so the cpu part can be just deleted and not have a stub with metal-specific implementation details like mxu

CC-Yeh: Didn't know we have private Kernel, thx

uuuvn (approved):

PR #591, opened 2026-07-08
Support Weaver Config

CC-Yeh:


CC-Yeh:
@uuuvn This could be closed? I think you mentioned we want to do it in other ways

CC-Yeh:
in favor of #611

PR #593, opened 2026-07-08
wasm downloader part 1: kiban

agolokoz:
crate with cross-platform OS primitives


agolokoz:
@codex review

crates/kiban/src/fs/opfs/directory_handle.rs
```diff
@@ -0,0 +1,108 @@
...
+        loop {
+            let step = JsFuture::from(iterator.next()?).await?;
+            if Reflect::get(&step, &JsValue::from_str("done"))?.as_bool().unwrap_or(true) {
+                break;
+            }
+            // Each value is a `[name, handle]` pair.
+            let pair = Array::from(&Reflect::get(&step, &JsValue::from_str("value"))?);
+            let name = pair.get(0).as_string().ok_or_else(|| JsValue::from_str("directory entry has no name"))?;
+            let handle = pair.get(1);
+            let entry = if handle.has_type::<FileSystemFileHandle>() {
```
eugenebokhan, line 97: can we use serde for decoding instead of manual conversion to string and parsing?

crates/kiban/src/fs/opfs/writable_file_stream.rs
```diff
@@ -0,0 +1,117 @@
...
+/// [`FileSystemWritableFileStream.write()`](https://developer.mozilla.org/en-US/docs/Web/API/FileSystemWritableFileStream/write).
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum WriteCommandType {
+    /// Write [`data`](WriteParams::data) at [`position`](WriteParams::position).
+    Write,
+    /// Move the cursor to [`position`](WriteParams::position).
+    Seek,
+    /// Resize the file to [`size`](WriteParams::size) bytes.
+    Truncate,
+}
```
eugenebokhan, line 20: can we use enums with associated values?

crates/kiban/src/fs/asyn.rs
```diff
@@ -0,0 +1,137 @@
...
+    #[cfg(target_family = "wasm")]
+    return super::asyn_opfs::dir_create_all(path.as_ref().to_str().unwrap()).await;
+
+    #[cfg(not(target_family = "wasm"))]
+    tokio::fs::create_dir_all(path).await
+}
+
+pub async fn exists(path: impl AsRef<Path>) -> bool {
+    try_exists(path.as_ref()).await.unwrap_or(false)
+}
```
eugenebokhan: feels like redundant

crates/kiban/src/fs/asyn_opfs.rs
```diff
@@ -0,0 +1,270 @@
...
+    get_file_handle_root(path, false).await.is_ok()
+}
+
+pub(crate) async fn get_dir_handle(
+    dir: &opfs::DirectoryHandle,
+    path: &str,
+    create: bool,
+) -> Result<opfs::DirectoryHandle, io::Error> {
+    dir.get_directory_handle(path, create).await.map_err(|err| js_value_to_io_error(&err, ErrorKind::Other))
+}
```
eugenebokhan, line 173: Can we migrate from a C style api to smth more like https://github.com/JohnSundell/Files ?

crates/kiban/src/fs/opfs/directory_handle.rs
agolokoz: It doesn't make sense. This code is more about working with js bridge objects and migrating to serde will not make the code smaller

crates/kiban/src/fs/opfs/writable_file_stream.rs
agolokoz: This enum just mirrors values of https://developer.mozilla.org/en-US/docs/Web/API/FileSystemWritableFileStream/write#type and serialized to js object

crates/kiban/src/fs/asyn_opfs.rs
agolokoz: The idea was to create simple API. Moreover this API will be hidden to users and object-oriented benefits will be lost. If create object-oriented filesystem API, then it's good to make it cross platform.
I don't think it's worth it to migrate this code to object-oriented API until file system access becomes performance bottleneck.

eugenebokhan (approved):

PR #604, opened 2026-07-14
Add LongRoPE config variant

sady4850:
adds `LongRoPEConfig` to `AnyRoPEConfig`, mirroring the lalamo side (trymirai/lalamo#325). Without it uzu cannot load Phi-3 / Phi-4 family models exported on long context, the config simply fails to deserialize.

HF selects `short_factor` or `long_factor` per forward pass, by `max(position_ids) + 1 > original_max_position_embeddings`. i did it this way first and got garbage on device.

your prefill is chunked: for a 5511 token prompt it is 6 calls, first 4 windows took short factors and last 2 took long. KV cache ends up with two different rotations and attention just collapses, output is fully broken

per sequence selection has the same problem: short prompt growing past 4096 during generation would flip factors on already short rotated cache. HF has it too, it just rarely hits it because prefill there is a single pass.

so the factor set must be constant for the whole sequence. Without plumbing new params through stream -> decoder -> transformer the only constant signal is the config, so I select by `max_sequence_length > original_context_length`. lalamo uses the same rule, export and engine agree

as a result artifact exported on long context uses long factors also for short prompts. To get HF behaviour the sequence length has to be plumbed into `precalculate` and pinned for the lifetime of the sequence. this is a bigger change


uuuvn (changes requested):
Responded in lalamo pr because the problem is in the that part: https://github.com/trymirai/lalamo/pull/325#issuecomment-4967798364

uuuvn (approved):

uuuvn:
Merged (once ci passes), thanks!

PR #629, opened 2026-07-21
Constants

eugenebokhan:


crates/backend-uzu/build/metal/gpu_types/gpu_type_constant.rs
```diff
@@ -0,0 +1,26 @@
...
+        "u16" => Ok("uint16_t"),
+        "u32" => Ok("uint32_t"),
+        "u64" => Ok("uint64_t"),
+        "f32" => Ok("float"),
+        "bool" => Ok("bool"),
+        "usize" => Ok("size_t"),
+        "isize" => Ok("intptr_t"),
+        unknown => bail!("Unsupported GPU constant type: {unknown}"),
+    }
+}
```
CC-Yeh: Looks really similar to `r2c` function from crates/backend-uzu/build/metal/gpu_types/gpu_type_struct.rs

Can we find a way to resuse it?

CC-Yeh: also check crates/backend-uzu/build/metal/gpu_types/gpu_type_option_set.rs

CC-Yeh (changes requested):

CC-Yeh (commented):
1. Accept const-expressions, not just literals. GpuTypeConstant::parse takes only Lit::Int/Float/Bool and bail!s on everything else, so a constant defined in terms of another (A + 1) or via an associated const (u32::MAX) is a hard build failure — both common shapes for derived sizes and sentinels, and derived ones are the most valuable to generate since hand-writing the resolved value is exactly the drift this feature prevents. Constants are emitted in source order into one namespace, so the expression can pass through verbatim and still be valid Metal C++. Suggest a whitelist walk of the syn::Expr (Lit, Path to a const in the same file, Unary, Binary, Paren), still bailing on casts and method calls, plus a test pinning declaration order since correctness would then depend on it.

2. Fail loudly instead of silently skipping. pub(crate) and private consts are Visibility::Restricted/Inherited, so they miss the matches!(&item.vis, syn::Visibility::Public(_)) guard, fall through to _ => None, and generate nothing with no error or warning. activation_type.rs:3-4 already has const GELU_K0: f32 = 0.044715; and GELU_K1, sitting in a scanned file and duplicated by hand in metal/kernel/activation/activations.h:23 as constexpr float k0 = 0.044715f; — precisely the case this PR exists to fix, and it stays silently ungenerated. Suggest erroring on any const in a gpu_types file that isn't generated, so the omission surfaces instead of being invisible.

3. Select by attribute, not visibility. Enums and structs are selected by location — every one in gpu_types/ is parsed, with #[repr(C)] enforced (item_enum.rs:50, item_struct.rs:65) — while constants are uniquely selected by visibility. That forces widening a constant to pub purely to trigger codegen, making it crate-public API as a side effect, which is what this PR does to MXU_SIMDGROUP_BLOCK_K at gemm.rs:42. Either add #[gpu_constant] (which also resolves 2) or drop the visibility filter for consistency with enums and structs.

CC-Yeh (approved):

PR #630, opened 2026-07-21
Tool calls base

agolokoz:
Implemented tools registration and calling.

During development there were many model specific errors, that's why there are many changes in *.json files.


agolokoz:
@codex review

PR #633, opened 2026-07-23
Apply GPU constant generation to weaver

CC-Yeh:


uuuvn (approved):

CC-Yeh:
one hack applied here is u32::MAX -> !0, since I don't want to add a string parsing for MAX of all types...

PR #637, opened 2026-07-24
Weaver cleanup + speedup

CC-Yeh:
### Perf
- Prefix KV now written directly into the cache. Removes the per-layer scratch→Global blit. ~2% faster `build_prefix`.
- Caches store KV only, not QKV. Prefix queries are never attended against. −33% cache memory.
- Last prefix layer stops after norm → qkv → KV write. Its attention + MLP output was always dropped. Saves ~7–8 dispatches per `build_prefix`. A 1-layer Weaver builds its prefix with no attention at all.
- Residual adds fused into norm shortcut path. Same idiom as `transformer_layer.rs`. Blocks exchange deltas; each norm absorbs the pending one. `TensorAddSwapKernel` removed from Weaver. −2 dispatches per block.
- Bias add moved inside the norm kernel. Deletes the `WeaverNorm` adapter and its extra dispatch.

### API / naming
- `WeaverPrefixKvCache`, `WeaverNodeKvCache`, `TopKChildren`, `PrefixLayerOutput` replace raw tuples.
- Node capacity stored in the cache. No longer re-derived from buffer size in the hot loop.
- `AttentionLastQuery` → `AncestorAttention`. Names the mechanism.
- Metadata/tree lanes use `MetadataIdx` / `TreeIdx` enums, not positional byte offsets.
- `residual_logits`: the readout is a correction added to draft pool scores before top-k.
- Weaver is BF16-only. `data_type` threading replaced with a `DATA_TYPE` constant.
- Dead surface removed: unused `linear_config`, always-1 `lookahead_offset`, unreachable zero-lookahead branch.

### DIFF
--- a/crates/backend-uzu/src/backends/common/allocator/allocator.rs
+++ b/crates/backend-uzu/src/backends/common/allocator/allocator.rs
@@ -26,13 +26,6 @@ impl<B: Backend> Allocation<B> {
         self.range.len()
     }
 
-    pub fn split_at_mut(
-        &mut self,
-        byte_offset: usize,
-    ) -> (BufferRangeMut<'_, B::DenseBuffer>, BufferRangeMut<'_, B::DenseBuffer>) {
-        self.as_buffer_range_mut().split_at(byte_offset)
-    }
-
     pub fn as_slice_mut<T: NoUninit + AnyBitPattern>(&mut self) -> &mut [T] {
         let buffer_range = self.as_buffer_range_mut();
         let (buffer, range) = (buffer_range.buffer(), buffer_range.range());
--- a/crates/backend-uzu/src/backends/common/allocator/allocator.rs
+++ b/crates/backend-uzu/src/backends/common/allocator/allocator.rs
@@ -26,6 +26,13 @@ impl<B: Backend> Allocation<B> {
         self.range.len()
     }
 
+    pub fn split_at_mut(
+        &mut self,
+        byte_offset: usize,
+    ) -> (BufferRangeMut<'_, B::DenseBuffer>, BufferRangeMut<'_, B::DenseBuffer>) {
+        self.as_buffer_range_mut().split_at(byte_offset)
+    }
+
     pub fn as_slice_mut<T: NoUninit + AnyBitPattern>(&mut self) -> &mut [T] {
         let buffer_range = self.as_buffer_range_mut();
         let (buffer, range) = (buffer_range.buffer(), buffer_range.range());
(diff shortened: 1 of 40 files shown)


CC-Yeh:
@codex review

crates/backend-uzu/src/backends/common/allocator/allocator.rs
```diff
@@ -26,6 +26,13 @@ impl<B: Backend> Allocation<B> {
         self.range.len()
     }
 
+    pub fn split_at_mut(
+        &mut self,
+        byte_offset: usize,
+    ) -> (BufferRangeMut<'_, B::DenseBuffer>, BufferRangeMut<'_, B::DenseBuffer>) {
+        self.as_buffer_range_mut().split_at(byte_offset)
+    }
```
uuuvn: I don't think "as buffer range mut and split at" is a good helper to have. It does two unrelated things and also is named wrong, `split_at_mut` on allocation would imply splitting one &mut Allocation into two &mut Allocation which is nonsensical (can't return &mut to a temp object). Allocation split_at can technically exist but would require changing allocator code (currently can't split allocations), but I don't see why.

I actually have an idea about how to simplify a lot of things surrounding buffers, allocations, buffer ranges that would not require doing this, but for now I don't think factoring out `self.as_buffer_range_mut().split_at(byte_offset)` into a helper is good

uuuvn (changes requested):

crates/backend-uzu/src/backends/common/allocator/allocator.rs
CC-Yeh: Yeah I think some view apis like other libs such as `numpy` would be good

uuuvn (approved):

crates/backend-uzu/src/backends/common/buffer/arg.rs
```diff
@@ -30,6 +30,13 @@ pub trait BufferArgMut<'a, B: Backend> {
     fn into_parts(self) -> (&'a dyn Buffer<Backend = B>, usize, usize);
 }
 
+impl<'a, B: Backend, Buf: Buffer<Backend = B>> BufferArgMut<'a, B> for BufferRangeMut<'a, Buf> {
+    fn into_parts(self) -> (&'a dyn Buffer<Backend = B>, usize, usize) {
```
eugenebokhan, line 34: I would suggest replacing the tuple with a struct with concrete fields for clarity

CC-Yeh: I agree it will look better, but this convention also exists in other 8 `into_parts`, not sure if we want to improve them all

eugenebokhan (approved):
LGTM

crates/backend-uzu/src/backends/common/buffer/arg.rs
eugenebokhan: Maybe in a separate PR?

PR #641, opened 2026-07-25
Visibility/boundary cleanup

CC-Yeh:
Move weaver stuff from dflash_speculator.rs ro weaver.rs

### DIFF
--- a/crates/backend-uzu/src/speculators/dflash_speculator.rs
+++ b/crates/backend-uzu/src/speculators/dflash_speculator.rs
@@ -9,18 +9,14 @@ use thiserror::Error;
 
 pub use crate::encodable_block::dflash::DFlashState;
 use crate::{
-    backends::common::{
-        Allocation, Backend, Encoder, Kernels,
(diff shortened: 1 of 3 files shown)


crates/backend-uzu/src/speculators/dflash_speculator.rs
```diff
@@ -205,40 +195,34 @@ impl<B: Backend> DFlashSpeculator<B> {
...
+                &mut encoder,
+            )?;
             let completed = encoder.end_encoding().submit().wait_until_completed().map_err(DFlashTreeError::Backend)?;
-            let slots = tree.copyout::<u32>();
-            drop(tree);
+            let nodes = tree.decode();
             drop(pool_ids);
             drop(pool_scores);
+            drop(draft_logits);
+            drop(draft_hidden);
```
eugenebokhan, line 223: why do manual drops?

eugenebokhan (approved):

crates/backend-uzu/src/speculators/dflash_speculator.rs
CC-Yeh: I think it's better lifetime management? so it won't be alive till the end of scope.

Yeah I will have at least 3-4 cleanup/perf PRs for dflash+weaver

PR #647, opened 2026-07-27
A8W8 bias/zp perf

CC-Yeh:
Integer divide in the zero-point path

- Slow: the correction recovered the weight column with scale_index / weight_groups_per_row — undoing a multiply for_each_line had performed two lines earlier. Ran 8× per weight group per thread (256× at k=2048), on hardware with no integer divide instruction, feeding an address computation that stalled behind it.
- Fixed: for_each_line passes the column it already computed. The 8-bit lookup additionally collapses to zero_points[scale_index], since the row stride is weight_groups_per_row at 8 bits.

Duplicate MMA for activation sums

- Slow: the correction needs Σₖ A_q[row,k], which an int8 MMA doesn't produce, so the kernel ran a second MMA of the matmul's own shape against a tile of ones — bias/sym measured exactly 2.00. The value depends only on activations, so all 24 threadgroups spanning n recomputed it; it landed in a TILES_M × TILES_N fragment, 32 registers holding 4 distinct rows; and it cost a full tensor op every 32 k-elements.
- Fixed: the quantize kernel emits it with one simd_sum over codes already in registers. The GEMM applies the correction as the rank-1 update it always was — once per weight group instead of once per activation chunk.


CC-Yeh:
**Speedup from the fixes (before / after, >1.00 = faster)**

| m | n3072 k2048 bias | zp | n1024 k3584 bias | zp |
|--:|--:|--:|--:|--:|
| 16 | 1.13x | 1.39x | 1.11x | 1.33x |
| 32 | 1.19x | 1.53x | 1.15x | 1.39x |
| 64 | 1.50x | 1.63x | 1.36x | 1.48x |
| 128 | 1.58x | 1.72x | 1.41x | 1.55x |
| 256 | 1.56x | 1.65x | 1.54x | 1.53x |
| 512 | 1.53x | 1.69x | 1.52x | 1.61x |
| 1024 | 1.49x | 1.64x | 1.48x | 1.61x |
| 2048 | 1.41x | 1.56x | 1.43x | 1.58x |

**a8w8 vs a16w8 after the fixes, GEMM only (>1.00 = a8w8 faster)**

| m | n3072 k2048 bias | zp | n1024 k3584 bias | zp |
|--:|--:|--:|--:|--:|
| 16 | 0.76x | 0.77x | 1.04x | 1.05x |
| 32 | 0.95x | 0.94x | 1.17x | 1.17x |
| 64 | 0.83x | 0.82x | 0.97x | 0.98x |
| 128 | 0.82x | 0.80x | 0.85x | 0.85x |
| 256 | 1.09x | 1.13x | 1.22x | 1.26x |
| 512 | 1.00x | 1.07x | 1.07x | 1.11x |
| 1024 | 1.11x | 1.20x | 0.95x | 1.00x |
| 2048 | 1.06x | 1.11x | 0.87x | 0.94x |

**Same, before the fixes, for reference**

| m | n3072 k2048 bias | zp | n1024 k3584 bias | zp |
|--:|--:|--:|--:|--:|
| 16 | 0.67x | 0.55x | 0.93x | 0.78x |
| 32 | 0.79x | 0.60x | 1.01x | 0.83x |
| 64 | 0.55x | 0.50x | 0.71x | 0.66x |
| 128 | 0.51x | 0.46x | 0.60x | 0.54x |
| 256 | 0.70x | 0.68x | 0.78x | 0.81x |
| 512 | 0.65x | 0.64x | 0.70x | 0.68x |
| 1024 | 0.74x | 0.73x | 0.64x | 0.62x |
| 2048 | 0.75x | 0.71x | 0.60x | 0.59x |

eugenebokhan (approved):

PR #650, opened 2026-07-27
Fix keisoku ios build (fails release)

uuuvn:


uuuvn:
(kimi slop, but looks like the right fix)

PR #669, opened 2026-08-02
Fix swapped N/K bounds in quantized block safe loaders

CC-Yeh:
`src_tile_dim` is (valid columns, valid rows) in the loader's own frame at both orientations


CC-Yeh:
@codex review

uuuvn (approved):

PR #678, opened 2026-08-05
Sleep tool in cli

uuuvn:


crates/cli/src/interactive/sessions/chat.rs
```diff
@@ -240,3 +241,16 @@ fn chat_state_mut(state: &mut ApplicationState) -> Option<&mut ChatSessionState>
 fn get_current_date_time() -> String {
     Local::now().to_rfc3339()
 }
+
+/// Sleeps for the given number of seconds before responding, at most 60 seconds.
+#[uzu_tool_function]
+async fn sleep(
+    /// Number of seconds to sleep, at most 60.
+    seconds: f64
```
agolokoz, line 249: How about milliseconds for more flexibility?

agolokoz (approved):

crates/cli/src/interactive/sessions/chat.rs
uuuvn: It's float, 0.1, 0.05, etc is possible. I wonder if llms might be better at handling and specifying milliseconds though...

PR #685, opened 2026-08-05
Weaver Update

eugenebokhan:
Replaced gelu with swiglu and positional embeddings with rope fused into weaver's ancestor attention following [lalamo](https://github.com/trymirai/lalamo/pull/333) changes


crates/backend-uzu/src/backends/metal/kernel/weaver/weaver_rope_qkv.metal
```diff
@@ -0,0 +1,45 @@
+#include <metal_stdlib>
+#include "../common/dsl.h"
+#include "weaver_frontier.h"
+
+using namespace metal;
+
+template <typename ElementT, typename RopeT>
+VARIANTS(ElementT, bfloat)
+VARIANTS(RopeT, float)
+PUBLIC KERNEL(WeaverRopeQkv)(
```
uuuvn: Is there a good reason to not use attention prepare instead of adding this?

eugenebokhan: already fused into attn

CC-Yeh (approved):

PR #687, opened 2026-08-06
Apply A8W4/8 optimizations

CC-Yeh:
## Optimizations

- Apply each optimization only where it wins; every one is gated by a flag dispatch sets from measurements
- Per-shape tiling and split-k targets: small m at decode doesn't produce enough threadgroups to fill the GPU otherwise
- Stage weight scales in threadgroup memory so simdgroups reuse them instead of each rebuilding the same values; off on big compute-bound tiles where the barrier costs more than the reuse saves
- Hoist operand addressing out of the k-loop; off on sym 128x128 where registers matter more than ALU
- Grouped drain: accumulate int32 products across a scale group and apply scales once per group instead of per chunk
- Independent activation/weight group sizes (act128/w64), with compact sidecars at true group granularity

## Redesign

- Rebuilt the MXU kernels into separate pieces: operand descriptors (what the data is), storage (the pointers), cursors (code loading), caches (scale/correction residency), and schedules (the loop)
- A new optimization lands as one flag or one specialization instead of another branch tangled into a monolithic kernel


CC-Yeh:
Todo:
- tune on M5/A18
- dispatch cleanup

crates/backend-uzu/src/backends/common/kernel/matmul/matmul_a.rs
```diff
@@ -9,6 +9,7 @@ pub enum MatmulA<'a, B: Backend> {
         values: &'a Allocation<B>,
         scales: &'a Allocation<B>,
         group_sums: Option<&'a Allocation<B>>,
+        activation_scale_group_size: u32,
```
eugenebokhan: I believe matmul should not know anything about activation

CC-Yeh: yeah, should've removed the activation_scale_ prefix

CC-Yeh:
@codex review

CC-Yeh:
Dispatch heuristics (host-side)
- Split-k target 256 (int8) — deeper K-splitting for decode; −8..−17% at m16-512.
- Int8 MXU tiling table — m/n-keyed tile selection (16x128 at m≤16, 32x64 at m<64, 128x128 at m≥512); −15..−43% at m16; with split-k 256 lands 1.2-2.0× decode together.
- Split-k 1024 on Tile32x64 (W8) — the decode tree-verify band (m17-63); −4.5..−6.2%.
- K-block floor on split-k — caps splits at one tile-block of K; −10% at k3584; gated to the int8 path.

Scale/metadata handling (kernel)
- Scale-line staging (TG) — all simdgroups share one cooperatively-filled scale+correction slab instead of each rebuilding register caches per weight group; helped asym most (−12..−29% at m64-256, up to −65% long-K) because the correction rebuild was the dominant repeated work.
- Double-buffer + prefetch — fill group g+1 while computing g, one barrier per group; turned staging's sym-128×128 penalty from +19-26% to half, improved small-m wins to −20%.
- Staging exclusions — sym-gs32 never stages, small-W4-decode never stages, sym-128×128 never stages (forcing it costs +31-41%: the barrier breaks load/compute overlap on that tile).
- Runtime min-groups guard (stage_scale_lines FC) — staging off when a Tile32x64 split covers <6 weight groups (those cells lost 13-15% staged); tunable in kernel.rs without touching shaders.
- Hoisted operand addressing — precompute scale/pointer strides outside the k-loop, advance incrementally; asym −3.3..−6.4% at m≥512. Gated OFF for sym-128×128 (non-hoisted recomputes addresses per group, saving registers there: −24% at m2048).

Drain (kernel)
- Grouped drain — accumulate int32 products across all chunks sharing the same scales (span = min of the two group sizes), sweep the accumulator once per span instead of per 32-chunk; helped decode most (m32 was ~20% behind without it) because the per-chunk f32 sweep dominates instruction issue at small m. Zero extra registers.

Quantization config
- Independent activation/weight group sizes (act128/w64 shipped) — coarser activation scales = 4× less sidecar traffic + cheaper prep; e2e −14.5% ms/fwd, +63% prompt t/s at g128 (measured coupled).
- Compact per-block group sums — fixed the H1 over-correction bug and lets any consumer weight-group granularity reconstruct its correction sums.

Threadgroup memory
- Guard-aware TG allocation — when the staging guard is off, the dispatch binds a minimal slab instead of ~1.5 KB; recovers occupancy on guarded-off cells.

Excluded (measured dead): zp-fusion (slower), naive int4→int8 unpack (60/60 dead), fused A8 epilogue (superseded by grouped drain), TG-staged weights, software pipelining, scratch-int8 repack (2× memory), per-row scales (deferred).

CC-Yeh:
@codex review

CC-Yeh:
@codex review

eugenebokhan (approved):

PR #690, opened 2026-08-07
Cleanup dflash

uuuvn:
Lalamo side: https://github.com/trymirai/lalamo/pull/335

[convert_dflash_checkpoint.py](https://github.com/user-attachments/files/30835194/convert_dflash_checkpoint.py)


eugenebokhan (approved):

crates/backend-uzu/src/encodable_block/weaver_layer.rs
```diff
@@ -0,0 +1,206 @@
...
+    pub num_heads: usize,
+    pub head_dim: usize,
+    pub max_depth: usize,
+}
+
+impl<B: Backend> WeaverLayer<B> {
+    pub fn new(
+        context: &B::Context,
+        config: &WeaverConfig,
+        add_to_residual: bool,
```
eugenebokhan, line 55: why not shortcut enum directly?

crates/backend-uzu/src/encodable_block/dflash.rs
```diff
@@ -90,37 +82,32 @@ pub enum DFlashDraftNewError<B: Backend> {
...
 
-impl<B: Backend> DFlashDraft<B> {
+impl<B: Backend> DFlash<B> {
     pub fn new(
         context: &B::Context,
         config: &DFlashDraftConfig,
         parameter_tree: &ParameterTree<B>,
         data_type: DataType,
-    ) -> Result<Self, DFlashDraftNewError<B>> {
+    ) -> Result<Self, DFlashNewError<B>> {
```
eugenebokhan, line 107: what's the point of NewError?

crates/backend-uzu/src/encodable_block/weaver_layer.rs
uuuvn: image

This stuff is just moved out of weaver.rs, this cleanup is focused on dflash, next one will be on weaver (likely unifying layer too so "weaver layer" nonsense won't be a thing anymore)

crates/backend-uzu/src/encodable_block/dflash.rs
```diff
@@ -275,11 +265,17 @@ impl<B: Backend> DFlashDraft<B> {
...
-        for ((layer, attention_state), key_value) in
+        for ((layer, mixer_state), key_value) in
             self.layers.iter().zip(state.layer_states.iter_mut()).zip(layer_key_values)
         {
-            attention_state.prepare(state.context_length, num_tokens, encoder.context())?;
-            layer.attention.append_projected_kv(key_value, &rope, num_tokens, attention_state, encoder)?;
+            mixer_state.prepare(state.context_length, num_tokens, encoder.context())?;
+            let attention = (layer.mixer.as_ref() as &dyn Any)
+                .downcast_ref::<Attention<B>>()
+                .expect("DFlash draft layers must use attention mixers");
```
eugenebokhan, line 274: either assert or do `map_err`

crates/backend-uzu/src/encodable_block/dflash.rs
```diff
@@ -289,17 +285,28 @@ impl<B: Backend> DFlashDraft<B> {
...
+        encoder.push_debug_group("dflash draft");
+
         let batch_dim = self.block_size;
         assert!(
             state.context_length + batch_dim <= self.max_context_length,
             "DFlash block positions exceed configured RoPE capacity"
         );
+
+        let mut tokens = vec![self.mask_token_id; batch_dim];
+        tokens[0] = target_output_token;
```
eugenebokhan, line 305: ai slope

crates/backend-uzu/src/encodable_block/dflash.rs
```diff
@@ -289,17 +285,28 @@ impl<B: Backend> DFlashDraft<B> {
...
+        let token_ids = encoder.allocate_constant_from_slice(&tokens).map_err(DFlashEncodeError::Backend)?;
+
```
eugenebokhan, line 307: why empty line

crates/backend-uzu/src/encodable_block/transformer_layer.rs
```diff
@@ -50,7 +50,9 @@ pub struct TransformerLayer<B: Backend> {
 impl<B: Backend> TransformerLayer<B> {
     pub fn new(
         context: &B::Context,
-        transformer_config: &TransformerConfig,
+        model_dim: usize,
+        hidden_dim: usize,
+        num_layers: usize,
```
eugenebokhan, line 55: why? seems like more loc for no reason

crates/backend-uzu/src/encodable_block/dflash.rs
uuuvn: Because new/encode have different potential errors (like parameter loader error is just new, but can never happen in encode)

crates/backend-uzu/src/encodable_block/dflash.rs
uuuvn: Proper error enum here is overkill because this is an internal invariant (all dflash layers are attention), assert will just be more code for exactly the same effect, I think unwrap/expect is the correct way to handle this stuff

crates/backend-uzu/src/encodable_block/dflash.rs
uuuvn: image

This was moved, but I also don't see what's wrong with it. chain once repeat collect just feels way more verbose than "all mask, but then first is target output"

crates/backend-uzu/src/encodable_block/dflash.rs
uuuvn: Yeah agree I think `let token_embeddings = target_embedding.encode_lookup(&token_ids, batch_dim, encoder)?;` should be the part of the block

uuuvn: Will fix in the followup cleanup prs

crates/backend-uzu/src/encodable_block/transformer_layer.rs
uuuvn: Transformer Layer was previously assuming it's a child of Transformer that has TransformerConfig, but DFlash doesn't have TransformerConfig and I think it's better to make transformer layer take these values separately than construct a fake TransformerConfig in DFlash.

PR #698, opened 2026-08-11
GPT-OSS MXFP4 MoE

xhjkl:
Dear Mirai team!

Thank you for making Uzu. I am building a CLI called `please`, whose current model support is centered on `gpt-oss`, and I would like to contribute that model path upstream.

This PR adds native GPT-OSS MXFP4 mixture-of-experts execution. Lalamo exports remain packed instead of expanding the large expert matrices to BF16, while Uzu's existing full-precision MoE path remains unchanged.

## Artifact contract

The official GPT-OSS checkpoint stores its fused projection as adjacent `[gate, up]` rows with group-32 MXFP4. The approved update in [Lalamo #339](https://github.com/trymirai/lalamo/pull/339) canonicalizes that representation during import:

- W13 rows become `[all up, all gate]`.
- Each original 32-value W13 block becomes two 16-value groups with the original E8M0 scale duplicated, so the represented values do not change.
- W2 remains ordinary group-32 MXFP4.
- The exported `MicrofloatSpec` fully describes the result; no GPT-OSS-specific layout mode is required.

Uzu validates that exact pair of specs (group-16 W13 and group-32 W2), keeps packed storage coupled to the MoE execution path, and dispatches dedicated decode and prefill kernels.

## Correctness

The branch is checked at several boundaries:

- Captured real GPT-OSS blocks, scales, biases, and global scales match a scalar dense reference.
- A multigroup fixture covers group-16 W13, group-32 W2, decode, and a 33-row prefill crossing the scheduler boundary.
- The block loader executes a Lalamo-shaped safetensors artifact and rejects mixed or unsupported expert specs.
- The focused MXFP4 suite passes 5 tests with 1 performance-only test ignored.
- The broader MoE suite passes 50 tests with 9 performance/production-scale tests ignored.

## Kernel work

The Metal kernels and host scheduling are adapted from the working GPT-OSS path in `llama.cpp` and expressed through Uzu's existing encoder, allocation, routing, and kernel abstractions. The branch also retains correctness-gated MXFP4-local improvements developed during testing: threadgroup lookup tables, vectorized prefill loads, and tuned decode/prefill tile geometry. It does not include the unrelated dense-Q8 work or any experimental packed layout.

The implementation is currently limited to Metal MoE experts using the canonical GPT-OSS combination of 4-bit group-16 W13 and group-32 W2. A generic arbitrary-group-size microfloat matmul API, NVFP4 kernels, and the suggested M5 integer-in-SRAM path remain follow-ups.

## Prerequisite

This PR consumes the artifact contract from [Lalamo #339](https://github.com/trymirai/lalamo/pull/339) or an equivalent.

### DIFF
--- a/crates/backend-uzu/src/backends/cpu/kernel/moe/counts_offsets_fused.rs
+++ b/crates/backend-uzu/src/backends/cpu/kernel/moe/counts_offsets_fused.rs
@@ -22,24 +22,37 @@ pub fn moe_counts_offsets_fused(
         return;
     }
 
-    // Phase 1: Count tokens per expert (histogram)
+    const SCATTER_BLOCK_SIZE: usize = 256;
+    const TILE_E: usize = 512;
+    let num_blocks = t.div_ceil(SCATTER_BLOCK_SIZE);
+    let num_tiles = e.div_ceil(TILE_E);
+
+    // Phase 1: Count tokens per expert and per scatter block.
     let mut counts = vec![0u32; e];
-    for ti in 0..t {
-        let base = ti * k;
-        for kk in 0..k {
-            let eid = unsafe { *topk_ids.add(base + kk) };
-            if eid >= 0 {
-                let ue = eid as usize;
-                if ue < e {
-                    counts[ue] += 1;
+    for block_id in 0..num_blocks {
+        let mut block_counts = vec![0u32; e];
+        let t_start = block_id * SCATTER_BLOCK_SIZE;
+        let t_end = (t_start + SCATTER_BLOCK_SIZE).min(t);
+        for ti in t_start..t_end {
+            let base = ti * k;
+            for kk in 0..k {
+                let eid = unsafe { *topk_ids.add(base + kk) };
+                if eid >= 0 {
+                    let ue = eid as usize;
+                    if ue < e {
+                        counts[ue] += 1;
+                        block_counts[ue] += 1;
+                    }
                 }
             }
         }
-    }
 
-    // Write partials (same as counts for single-threadgroup case)
-    for i in 0..e {
-        unsafe { *partials.add(i) = counts[i] };
+        for (expert_id, count) in block_counts.into_iter().enumerate() {
+            let tile_id = expert_id / TILE_E;
+            let tile_expert_id = expert_id % TILE_E;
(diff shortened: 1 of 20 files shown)


norpadon:
Hi! First of all, we appreciate your contributions, you are doing something we wanted to do for a long time but haven't had time to work on! Amazing.

I want @CC-Yeh to take a closer look at the kernels. Ideally we probably want a generic matmul API which supports arbitrary group sizes and both mxfp4 and nvfp4 modes, just like lalamo. It is fine if only mxfp4 shapes are implemented in the beginning.

There is a very important optimization opportunity here:

E2M1 can only represent integer and half-integer values: 0, +-0.5, +-1.0, +-1.5, +-2.0, +-3.0, +-4.0, +-6.0

This means that if we multiply them by 2, and adjust the scale factors accordingly, we can represent them as int8 values in SRAM, and take advantage of the accelerated MXU int8*int8 matmul path for prefill and speculative decoding on M5 chips. The latest metal4.1 also supports block-scaled microfloats directly, and may already do this exact optimizaton (although we are not sure, and it is not very stable yet).

norpadon:
As a reward for your hard work we will cook up a SotA 4-bit and 8-bit quantizations for dense layers for you once we stabilize the format 

cc @knyazer

CC-Yeh:
Hi @xhjkl ! Thanks for your contribution.

We're planning to revamp the current MoE kernels, so extending the old kernels may not align well with what we want the implementation to go.

The main issues with the current MoE implementation are:

- Limited reuse of the existing matmul infrastructure. Ideally, expert computation should use a shared GatheredGEMV/GEMM interface.
- Separate prefill and decode paths, even though routing is the same and only the matmul workload differs. 
- Too many intermediate routing, gather, and scatter steps.
- Duplicate Metal kernels and Rust dispatch code across paths and dtypes.

Speed, maintainability, and correctness are our priorities.

We’d like to address these more fundamentally, possibly with GatheredGEMM or routing/top-k fusion, but feel free to explore better approaches.

 I’d suggest breaking this into smaller, reviewable PRs and reusing existing code wherever possible.

xhjkl:
@CC-Yeh, thanks a lot for such a detailed feedback! Let's indeed break this change set into smaller, independently mergeable PRs. I'll do this shortly.

xhjkl:
Superseded by #725 and the subsequent stack.

PR #699, opened 2026-08-11
Execute fused QKVG attention projections

xhjkl:
Consumes Lalamo's fused QKVG projection directly. This eliminates the separate gate-projection dispatch and the gated-input scratch copy while retaining explicit packed-row strides for attention preparation, Q/K/V normalization, and post-attention sigmoid gating.

### Benchmarks

- **BF16:** decode improved from 169.48 to 186.74 tokens/s (**+10.18%**); long-prompt prefill throughput regressed 1.29%; greedy output was identical; peak memory increased by 4 MiB.
- **MLX 4-bit:** decode improved from 283.35 to 310.99 tokens/s (**+9.75%**); long-prompt prefill throughput regressed 0.49%; decode peak memory was unchanged; long-context peak memory decreased by 16 MiB.

### Validation

- `cargo test -p backend-uzu qkv_norm -- --nocapture`
- `cargo test -p backend-uzu attention_prepare -- --nocapture`
- `cargo test -p backend-uzu sigmoid_gate -- --nocapture`
- `cargo clippy -p backend-uzu --lib --tests -- -D warnings`

### Dependency

Requires [trymirai/lalamo#340](https://github.com/trymirai/lalamo/pull/340) and a Lalamo release containing it.


uuuvn (changes requested):
Thank you very much for taking a time to do this!

One thing is that we shouldn't name variables "qkv" when they are in fact "qkvg", other than that looks good, will merge after https://github.com/trymirai/lalamo/pull/340 is merged and a lalamo version is released with it.

crates/backend-uzu/src/config/token_mixer/attention.rs
```diff
@@ -7,7 +7,8 @@ use crate::config::{
 
 #[uzu_config(super::TokenMixerConfig)]
 pub struct AttentionConfig {
-    pub qkv_projection_config: LinearConfig,
+    /// Unified QKVG projection; ungated attention omits the G output segment.
```
uuuvn: Useless llm comment

crates/backend-uzu/src/encodable_block/mixer/attention/mod.rs
```diff
@@ -85,32 +85,32 @@ impl<B: Backend> Attention<B> {
...
-        let qkv_projection_tree = parameter_tree.subtree("qkv_projection")?;
+        let qkvg_projection_tree = parameter_tree.subtree("qkvg_projection")?;
         let qkv_dim = if let Some(num_kv_heads) = num_kv_heads {
             let kv_dim = num_kv_heads * head_dim;
             q_dim + kv_dim + kv_dim
         } else {
             q_dim
         };
-        let projection_dim = if config.has_gate {
+        let qkvg_dim = if config.has_gate {
```
uuuvn: Why change "projection" to "qkvg"?

crates/backend-uzu/src/encodable_block/mixer/attention/mod.rs
```diff
@@ -85,32 +85,32 @@ impl<B: Backend> Attention<B> {
...
-            &qkv_projection_tree,
+            &qkvg_projection_tree,
         )?;
 
         let query_norm_config = config.query_norm_config.clone();
         // TODO: Fix lalamo config, those two must be None if kv sharing.
         let key_norm_config = (!is_kv_sharing).then(|| config.key_norm_config.clone()).flatten();
         let value_norm_config = (!is_kv_sharing).then(|| config.value_norm_config()).flatten();
-        let packed_projection_heads = projection_dim / head_dim;
+        let qkvg_heads = qkvg_dim / head_dim;
```
uuuvn: There is no gate heads, `qkvg_heads` is wrong here

crates/backend-uzu/src/encodable_block/mixer/attention/mode.rs
```diff
@@ -19,23 +19,23 @@ use crate::{
     utils::maybe_mut::MaybeMut,
 };
 
-pub(super) struct LinearProjection<B: Backend> {
+pub(super) struct QKVGProjection<B: Backend> {
```
uuuvn: Why rename linear?

uuuvn (changes requested):

crates/backend-uzu/src/config/token_mixer/attention.rs
xhjkl: Ha, I wrote it myself 🤭 Just wanted to emphasize the "G" part may be skipped. But I'll drop this comment in the subsequent update if you feel that way.

crates/backend-uzu/src/encodable_block/mixer/attention/mode.rs
xhjkl: Purest misfire on my end, as I'm still getting ramped up with the codebase knowledge. Now fixed.

uuuvn (approved):
LGTM, will merge when lalamo part merges and releases, thanks

gearonixx:
@xhjkl 

In older versions, the kernel computed the offset internally

```rust
// older qkv_norm.rs kernel

let qkv_stride = total_heads * head_dim; 

let offset =
    (token_index * qkv_stride)
    + HEAD_START_OFFSET;
```

After these changes, the stride is passed explicitly as `input_row_stride`

```rust
self.encode_packed(
    key_value,
    batch_dim,
    0,
    2 * self.num_kv_heads, // this is the number of heads. should be the number of elements in the token
    encoder,
)
```

should be:

```diff
-        self.encode_packed(key_value, batch_dim, 0, 2 * self.num_kv_heads, encoder)
+        self.encode_packed(key_value, batch_dim, 0, 2 * self.num_kv_heads * self.head_dim, encoder)
```

`* self.head_dim` is missing

The kernel still computes:

```rust
let offset =
    (batch * input_row_stride) // batch * number_of_elements (not number of heads)
    + HEAD_START_OFFSET;
```


`input_row_stride` needs to be the number of elements per token

this path only happens when DFlash synchronizes accepted speculative tokens through `DFlash::encode_accept`

xhjkl:
Great catch, @gearonixx! Now fixed ✨

uuuvn:
We finally did everything on our side for the new model format, merged into the new format branch, will be in main soon, thanks for your contribution!

PR #700, opened 2026-08-11
Whisper log-mel frontend

xhjkl:
Adds an optional, host-side Whisper audio frontend to `backend-uzu`.

The new `backend-uzu/audio` feature converts a normalized 16 kHz `PcmBatch` into the fixed-shape, frame-major `f16` log-mel input expected by Whisper: `[3_000, 80]` or `[3_000, 128]`.

This keeps the preprocessing contract independently testable before adding model execution. It intentionally does not include Whisper inference, model acquisition, sessions, or bindings.

<details>
<summary>What changed</summary>

- Adds `backend_uzu::audio::whisper_log_mel_spectrogram`.
- Validates sample rate, batch shape, channel interleaving, duration, finiteness, and normalized PCM range.
- Averages interleaved channels and zero-pads short input to Whisper’s 30-second window.
- Implements Whisper’s centered periodic-Hann STFT, Slaney mel filters, and dynamic-range normalization.
- Returns a typed `WhisperLogMelSpectrogram` with an explicit `[frame, mel_bin]` layout.
- Supports both 80-bin and 128-bin Whisper checkpoints.
- Keeps `rustfft` optional behind the new `audio` feature.
- Enables the feature in macOS CI without affecting default Linux, WASI, or wasm builds.

`PcmBatch` does not carry channel-layout metadata, so multichannel input uses an unweighted arithmetic mean. Callers with nonuniform channel layouts should downmix before invoking the transform.

</details>

## Correctness

Parity fixtures were captured from a pinned OpenAI Whisper revision using a real excerpt from its JFK test audio. The repository includes the source hashes and exact reproduction script.

The tests compare every mel bin at frames spanning the signal, signal/padding boundary, and padded region for both 80- and 128-bin outputs. They also cover invalid PCM contracts, the inclusive 30-second boundary, stereo downmixing, output shape, and non-finite reference protection.

Tests requiring the repository’s external model/tokenizer bundle were left to CI.


uuuvn:
Not something we want to have in uzu yet

PR #702, opened 2026-08-11
Cli fixes

agolokoz:
1. Average stats for sessions with tool calls
2. Another small ui fixes


agolokoz:
@codex review

agolokoz:
@codex review

uuuvn (changes requested):
This is not the right fix, cli shouldn't do it's own config through uzu, and uzu won't have any persistent "settings" at all after refactor of high level crates. Cli should do $XDG_CONFIG_HOME/mirai/config.toml

agolokoz:
> This is not the right fix, cli shouldn't do it's own config through uzu, and uzu won't have any persistent "settings" at all after refactor of high level crates. Cli should do $XDG_CONFIG_HOME/mirai/config.toml

Preferences refactoring will be in separate PR

agolokoz:
@codex review

eugenebokhan (approved):

uuuvn (approved):
I don't fully understand what this code does but all of that is going to be rewritten in the big high level crate refactor so doesn't matter I guess

PR #703, opened 2026-08-12
 Fix GDN state precision across update paths

CC-Yeh:
- Preserve F32 normalized K for persistent-state updates while retaining activation-dtype K for tensor-core computations.
  - Apply consistent ordinary exp/log/rsqrt math across recurrent, chunked-prefill, and tree-verification kernels.
  - Remove explicit metal::precise overrides and precise SiLU handling.
  - Keep the strict out.metal MXU path.
  - Add coverage for the F32 state-key preparation and state advancement paths.


CC-Yeh:
Does not see any observable difference in acceptance rate from this fix

eugenebokhan (approved):

CC-Yeh:
within accuracy tolerance, no impact on acceptance rate

PR #708, opened 2026-08-12
Traces

eugenebokhan:
WIP


crates/backend-uzu/src/backends/common/encoder.rs
```diff
@@ -121,7 +127,7 @@ impl<'encoding, B: Backend> Encoder<'encoding, B> {
         let src_range = resolve_copy_range(src_range, src_buffer_range.range().len(), "source");
         let dst_range = resolve_copy_range(dst_range, dst_buffer_range.range().len(), "destination");
         let byte_len = src_range.len();
-        assert_eq!(byte_len, dst_range.len(), "copy range lengths must match");
+        assert_eq!(byte_len as usize, dst_range.len() as usize, "copy range lengths must match");
```
eugenebokhan: will remove

ry2009:
@uuvn tracing audit notes before using Lalamo as the verifier for this PR:

The broad shape is right: Uzu exports safetensors, the comparator stays on the Lalamo side, and the CLI tokenization metadata includes the resolved system prompt plus `bos_token`/`eos_token`.

Concrete blockers I would fix before trusting the round trip:
1. `external/lalamo-worktrees` is a symlink to `/Users/eugenebokhan/Developer/trymirai/lalamo-worktrees/`; this looks accidental and should not be in the PR.
2. Trace token ids are accepted as `u64` and cast with `as` to `u32` for runtime input and `i32` for `activation_trace.token_ids`. Values outside the Lalamo trace schema would silently wrap/truncate. This should use checked conversion and fail loudly.
3. Duplicate trace paths are only guarded with `debug_assert!` in the recorder, so release builds can overwrite an earlier tensor in the map and emit a misleading trace. This should be a hard error/assert in all builds.

After those are fixed, the acceptance test I care about is: export a real Uzu trace, replay the saved `activation_trace.token_ids` and `activation_trace.token_positions` through Lalamo, then pass Lalamo `compare-traces` with no Uzu-side comparator.

ry2009:
went through this again on ed05659 -- the tap struct rewrite is a big improvement imo. paths falling out of the type tree instead of getting assembled into a string keyed hashmap means duplicate/misplaced tensors just can't happen anymore, which is better than the assert I was going to suggest. good that the macros in utils/trace.rs are gone and Encoder isn't carrying a recorder around either, and symlink cleanup.

also checked the wire format survived the rewrite and it did -- `activation_trace.token_ids`, `activation_trace.token_positions` and `activation_trace.layer_results.N.activation_trace.*` all still line up with the lalamo side name for name, classifier double logits included.

stuff I still see open:

1. token ids are u64 and reach the trace through unchecked `as` casts -- `engine/language_model/trace.rs:78,100,101` and `engine/classifier_model.rs:116,186,193,194`. anything >= 2^31 wraps to a negative int32 in `activation_trace.token_ids` and the trace still writes successfully, so you get a silently wrong trace which is the one failure mode we really can't have here. `(0..token_count as i32)` is also casting the length rather than the elements. narrowing the trace path to u32 + `i32::try_from` covers it, trie stays u64 since thats engine wide so it just needs one widen at the `TrieNode::flat` call.

2. no vocab size check on the ids: an out of range id indexes past the embedding table and still writes a successful trace, lalamo only catches it when it reads the file. checking against `Embedding::vocab_size()` at the source catches it where it happens, and makes the int32 check in 1 redundant anyway since any real vocab fits in i32.

3. metadata is optional:`trace/record.rs:16,29,50` thread `Option<HashMap<String, String>>` so you can emit a traces.safetensors with no tokenization metadata at all and no error. if the tokenization trace is part of the contract it prob wants to be a required typed struct.

4. `crates/cli/src/trace.rs:50,54` hardcode `add_generation_prompt: true` and `enable_thinking: true`, and the config gets walked as untyped serde_json::Value. bigger version of this is that the cli reimplements the prompt pipeline instead of going through the production renderer -- real chat goes through hanashi's Renderer with the shoji HanashiConfig template, role canonization, configurable preamble key, trim_start. the forward pass itself is the real engine path so the divergence is just template source + request assembly, but it means compare-traces can pass while the traced tokenization isn't what an actual session feeds the model. a template with leading whitespace diverges immediately.

5. classifier traces get chat rendered: `record.rs:20` dispatches `ModelSpecialization::Classification` but `cli/src/trace.rs` has no specialization branch. the classification runtime tokenizes with add_generation_prompt=false, enable_thinking=false and no system prompt (`hanashi/src/classification/encoding.rs`), so a classifier trace is validating a path that doesn't exist in prod. either dispatch on it or bail until its supported.

6. `crates/cli/Cargo.toml:36,37,39:  hanashi/minijinja/tokenizers are unconditional deps but only src/trace.rs uses them, could be optional and pulled in by the trace feature.

7. `crates/backend-uzu/src/lib.rs:20`: ChatCodecConfig re-export isn't gated so its public at the crate root in every build, only the trace cli consumes it.

one more thing thats not really either PR's fault but might hurt on the first real comparison: post_layer_scalar gets folded into different tensors on each side. uzu applies it inside pre_mlp_norm's shortcut write and inside post_mlp_norm, so traced `mlp_inputs` = scalar*(inputs+mixer) and traced `post_mlp_norm` = scalar*norm(mlp). lalamo traces both unscaled and only applies the scalar to `outputs`. final outputs agree bc rms/layernorm are scale invariant so its invisible on most models, but anything with has_post_layer_scalar and no PLE will fail those two tensors by exactly that scalar. we should pick a convention before the acceptance run or the first compare is gonna look like a kernel bug.

on the lalamo side im rebuilding the tracer branch as a clean PR off main so theres actually something to verify against -- will link it once its up. happy to send 1 as a patch against ed05659 rather than pushing to your branch if thats easier.

PR #713, opened 2026-08-13
Disable metallib compression

uuuvn:


uuuvn:
Doesn't appear to be the actual trigger, flaky

PR #714, opened 2026-08-14
Cli reasoning effort

agolokoz:


agolokoz:
@codex review

crates/cli/src/interactive/components/application.rs
```diff
@@ -38,6 +39,8 @@ pub struct ModelState {
 
 pub struct ApplicationState {
     preferences: Preferences,
+    /// Session-scoped thinking override from the command line; never persisted.
```
eugenebokhan, line 42: can be removed

crates/cli/src/interactive/components/application.rs
```diff
@@ -38,6 +39,8 @@ pub struct ModelState {
...
+    thinking_override: Option<ThinkingPreference>,
```
eugenebokhan, line 43: why do override instead of updating the prefs?

agolokoz:
```
cli --reasoning-effort high
``` 
session with high effort

```
cli
``` 
session with high effort (saved from previous run), but user didn't specify it, wrong behaviour

crates/cli/src/interactive/components/application.rs
agolokoz: I think this comment answers the obvious question about why `thinking_override: Option<ThinkingPreference>` exists. I would like to keep this comment.

eugenebokhan (approved):

crates/backend-uzu/src/backends/metal/kernel/attention.rs
```diff
@@ -85,7 +85,7 @@ impl AttentionGemmCore<Metal> for AttentionGemmMetalCore {
         arguments: &AttentionCoreNewArguments,
         _context: &MetalContext,
     ) -> Result<bool, MetalError> {
-        Ok(matches!(arguments.head_dim, 64 | 128 | 256))
+        Ok(matches!(arguments.head_dim, 64 | 128) || arguments.head_dim == 256 && arguments.data_type != DataType::BF16)
```
agolokoz: Without this line inference with `cli --reasoning-effort disabled` shows error 
"Chat error: Backend error: Decoder error: Backend error: Cannot create pipeline state: Compilation failed due to an interrupted connection: XPC_ERROR_CONNECTION_INTERRUPTED. This error occurred after multiple retries."

Codex suggested such fix with description
"The CLI’s registered tools lengthen the prompt and select an unaligned BF16 AttentionGemm specialization with head_dim=256; Apple’s Metal compiler crashes compiling that specialization. Quick-start happened to select the aligned variant."

PR #715, opened 2026-08-14
Reliable MoE Routing Beyond 256 Tokens

xhjkl:
MoE routing processes tokens in 256-token blocks, but its counting kernel emitted one global expert histogram while scatter expected one histogram per block. Beyond 256 tokens, later routes could overlap existing expert rows or address rows outside the routed output.

This change gives every token block its own expert histogram while preserving global expert offsets, so longer prefills retain a one-to-one mapping between routed tokens and expert rows. It also corrects offset scanning above 128 experts. Routing decisions, expert computation, model formats, and public APIs remain unchanged. This issue was discovered during #698 but is independent of its MXFP4 work.

### DIFF
--- a/crates/backend-uzu/src/backends/metal/kernel/moe/counts_offsets_fused.metal
+++ b/crates/backend-uzu/src/backends/metal/kernel/moe/counts_offsets_fused.metal
@@ -35,32 +35,32 @@ static T threadgroup_raking_prefix_exclusive_sum(T value, threadgroup T* shared,
   return result;
 }
 
-#define BLOCK_SIZE 128
+#define THREADGROUP_SIZE 128
 #define SCATTER_BLOCK_SIZE 256
-#define TILE_E 512
+#define EXPERT_TILE_SIZE 512
 
 // Single-kernel fused: count all experts + scan to offsets
 // This kernel is launched with SINGLE threadgroup
 PUBLIC KERNEL(MoeCountsOffsetsFused)(
-    device const int* topk_ids,
-    device uint* offsets,   // output: exclusive scan [E+1]
-    device uint* sum_k_out, // output: total count [1]
-    device uint* partials,  // output: partials [num_blocks * num_tiles * TILE_E] (for block_bases)
-    constant uint& t_input,
-    constant uint& e_input,
-    constant uint& k_input,
-    threadgroup _atomic<uint> tg_hist[TILE_E],
-    threadgroup uint scan_shared[BLOCK_SIZE],
-    threadgroup uint reduce_shared[BLOCK_SIZE],
-    threadgroup uint counts_shared[TILE_E], // Cache global counts in threadgroup memory
-    threadgroup uint& carry,
+    device const int* selected_expert_ids,
+    device uint* expert_offsets,              // output: exclusive scan [expert_count + 1]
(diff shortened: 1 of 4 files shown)


crates/backend-uzu/src/backends/metal/kernel/moe/counts_offsets_fused.metal
```diff
@@ -35,104 +35,133 @@ static T threadgroup_raking_prefix_exclusive_sum(T value, threadgroup T* shared,
   return result;
 }
 
-#define BLOCK_SIZE 128
-#define TILE_E 512
+#define THREADGROUP_SIZE 128
+#define SCATTER_BLOCK_SIZE 256
+#define EXPERT_TILE_SIZE 512
```
CC-Yeh, line 40: Duplicated definitions, check backends/common/gpu_types

CC-Yeh (changes requested):
Thanks!

I'm confused about the current max routed exeprts is `512`, but the `EXPERT_TILE_SIZE` suggest it can go beyond 512.

```rust
        if moe_config.num_routed_experts > 512 {
            return Err(MoeBlockError::InvalidRoutedExpertCount);
        }
```

CC-Yeh:
Thanks for continuing to explore this.

I’d like to replace the current route-preparation pipeline rather than extend it. The counts, offsets, bucket construction, and permutation maps create several intermediate buffers and dependent dispatches before expert computation can begin.
This adds memory traffic and GPU bubbles, while making the path difficult to understand, review, and maintain.

At a high level, I think the MoE path should look like this:

  routes = router.route(input)
  output = experts.encode(input, routes)

routes should contain route-major expert IDs and routing weights. Expert execution could then follow:

 gathered gate/up projection
  → gated activation
  → gathered down projection
  → route-weighted reduction

The backend should consume TopK expert IDs directly. Decode and other small-batch workloads are generally memory-bound, so direct GatheredGEMV should avoid unnecessary preparation, intermediate memory traffic, and dependent dispatches.

For prefill, and potentially speculative decoding with enough routed rows, GatheredGEMM can reuse expert weights across rows and increase arithmetic intensity. Any route grouping should remain private to the backend and only be enabled when benchmarks show that it pays off.

Some promising fusion points are:
  - Router projection + TopK for small batches.
  - Gathered gate/up projection + gated activation.
  - Gathered down projection + routing-weight application and reduction.

Prefill and decode should share the same interface, with the backend selecting GatheredGEMV or GatheredGEMM based on the workload.

Given this direction:
  - I don’t think #715 is necessary.
  - The validation from #716 could be submitted independently.
  - For #717, I suggest simplifying the contract so it accepts TopK expert IDs directly and produces route-major output, without segmented offsets or permutation maps.

What do you think about this design? I’m also interested if you see a simpler or faster way to reach the same goal.

cc @eugenebokhan @uuuvn

xhjkl:
@CC-Yeh, thanks a lot for such a detailed review! I truly appreciate the time and effort you've put into spelling out the better direction. I think you have much more visibility into the domain area than I do — so I'd better defer the strategy to your taste.

At this point I think I'll need a bit of time to fully internalize the design you proposed, and then I'll come back to you with a new pack of PRs 🫶

PR #716, opened 2026-08-14
MoE Rejects Impossible Expert Counts

xhjkl:
MoE routing requires at least one routed expert, at least one active expert, and no more active experts than routed experts. Invalid configurations currently reach backend-specific routing behavior instead of failing consistently.

This change enforces those constraints when the full-precision MoE block is constructed, while preserving the supported limits of 512 routed experts and 128 active experts. It changes no routing kernels, expert computation, or buffer initialization.

### DIFF
--- a/crates/backend-uzu/src/backends/metal/kernel/moe/counts_offsets_fused.metal
+++ b/crates/backend-uzu/src/backends/metal/kernel/moe/counts_offsets_fused.metal
@@ -35,32 +35,32 @@ static T threadgroup_raking_prefix_exclusive_sum(T value, threadgroup T* shared,
   return result;
 }
 
-#define BLOCK_SIZE 128
+#define THREADGROUP_SIZE 128
 #define SCATTER_BLOCK_SIZE 256
-#define TILE_E 512
+#define EXPERT_TILE_SIZE 512
 
 // Single-kernel fused: count all experts + scan to offsets
 // This kernel is launched with SINGLE threadgroup
 PUBLIC KERNEL(MoeCountsOffsetsFused)(
-    device const int* topk_ids,
-    device uint* offsets,   // output: exclusive scan [E+1]
-    device uint* sum_k_out, // output: total count [1]
-    device uint* partials,  // output: partials [num_blocks * num_tiles * TILE_E] (for block_bases)
-    constant uint& t_input,
-    constant uint& e_input,
-    constant uint& k_input,
-    threadgroup _atomic<uint> tg_hist[TILE_E],
-    threadgroup uint scan_shared[BLOCK_SIZE],
-    threadgroup uint reduce_shared[BLOCK_SIZE],
-    threadgroup uint counts_shared[TILE_E], // Cache global counts in threadgroup memory
-    threadgroup uint& carry,
+    device const int* selected_expert_ids,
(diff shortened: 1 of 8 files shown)


xhjkl:
Apparently, I cannot stack my PRs properly unless I'm the repo owner.

This PR will show extra commits until its preceding PR is merged in, so I'm keeping it draft until that happens.

crates/backend-uzu/src/backends/metal/kernel/moe/counts_offsets_fused.metal
```diff
@@ -67,40 +68,51 @@ PUBLIC KERNEL(MoeCountsOffsetsFused)(
...
-        int eid = topk_ids[base + k];
-        if (eid >= 0) {
-          uint ue = uint(eid);
-          if (ue >= e0 && ue < e0 + tile_e) {
-            uint te = ue - e0;
-            atomic_fetch_add_explicit(&tg_hist[te], 1u, memory_order_relaxed);
+    // Scatter consumes one partial histogram per 256-token block.
+    for (uint block_id = 0; block_id < num_blocks; ++block_id) {
+      for (uint e = lid; e < tile_e; e += BLOCK_SIZE) {
+        atomic_store_explicit(&tg_hist[e], 0u, memory_order_relaxed);
```
eugenebokhan: why filling with zeros needs atomic?

crates/backend-uzu/src/backends/metal/kernel/moe/counts_offsets_fused.metal
```diff
@@ -67,40 +68,51 @@ PUBLIC KERNEL(MoeCountsOffsetsFused)(
...
+      const uint t_end = min(t_start + SCATTER_BLOCK_SIZE, t_input);
+      for (uint t = t_start + lid; t < t_end; t += BLOCK_SIZE) {
+        const uint base = t * k_input;
+        for (uint k = 0; k < k_input; ++k) {
+          int eid = topk_ids[base + k];
+          if (eid >= 0) {
+            uint ue = uint(eid);
+            if (ue >= e0 && ue < e0 + tile_e) {
+              uint te = ue - e0;
+              atomic_fetch_add_explicit(&tg_hist[te], 1u, memory_order_relaxed);
```
eugenebokhan: please, rewrite every variable name to full explicit and human-readable for clarity

eugenebokhan (changes requested):

crates/backend-uzu/src/backends/metal/kernel/moe/counts_offsets_fused.metal
xhjkl: The zeroing itself does not need atomic coordination; each thread initializes distinct cells. However, `tg_hist` must have atomic element type because multiple threads subsequently increment the same cells, and Metal does not permit an ordinary assignment to `_atomic<uint>`. I verified this by replacing the relaxed atomic store with a plain assignment in a minimal public reproduction; [Metal 4.0 rejects it here](https://github.com/xhjkl/firingrange/actions/runs/31942659315/job/95153965641#step:3:5). The relaxed atomic store is therefore just the legal initialization; the following barrier provides the inter-thread ordering before the increments begin.

PR #718, opened 2026-08-14
Deslope

eugenebokhan:


uuuvn:
This allows owners/&mut to modify fields like is_flat potentially violating struct invariants. I don't like that rust doesn't have a simple way to make a struct field pub-read-only without adding a getter function but I don't think the increase in complexity of tracking struct field invariants is worth the simplification here

PR #719, opened 2026-08-14
Cpu backend perf

Bo0oM:
Optimizations for the hot path of the CPU backend (linux/windows/x86-mac) plus a few changes to the common code that remove work from each dispatch on both backends.

Sampling. unified_sampling sorted the entire dictionary to take the prefix. Now, partial candidate selection via select_nth_unstable_by with an increase in sampling when a miss occurs on top-p. The order of candidates is defined by a strict complete order (value in descending order, index in ascending order), so the prefix matches the previous one by construction.

A worker pool has been added to CpuContext, and matmul is divided across the output columns. Quantization group constants (scale, zero-point, bias) and data type dispatching have been moved out of the inner loop; 4‑bit codes are read as a single u32 word for eight codes; the activation string is decoded once per dispatch, not per column.

Backend::NEEDS_BARRIERS — both backends have an empty encode_barrier, so tracking hazards and building Vec<Access> in the code generator are no longer performed. Debug groups in Metal do not create String and NSString unless debugging or capturing is enabled. Allocator pool numbers are returned to the free list — previously, the per-pool state grew with the number of pools ever created. FlatTrie::accept uses a flat index instead of a linear search by pointers.

I ran the benchmarks on M1 Max, Qwen3-0.6B-MLX-4bit, and there is a several‑fold speedup. I haven’t tested iOS (metal); the changes in metal/command_buffer.rs, metal/context.rs, and in the code generator have only been verified by reading. We need to run your benchmarks and measure the time on the device.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -241,6 +241,7 @@ opt-level = 2
 
 [profile.release]
 lto = "thin"
+codegen-units = 1
 
 [profile.test]
 opt-level = 2
--- a/crates/backend-uzu/build/metal/bindgen/arguments.rs
+++ b/crates/backend-uzu/build/metal/bindgen/arguments.rs
@@ -405,7 +405,9 @@ pub fn encode_accesses_call(arguments: &[ArgumentEmission]) -> TokenStream {
         quote! {}
     } else {
         quote! {
-            encoder.access(&[#(#access_expressions),*].into_iter().flatten().collect::<Vec<_>>());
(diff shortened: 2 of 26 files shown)


uuuvn:
CPU backend is mainly a reference for tests, it should be kept as simple as possible. This introduces a ton of complexity (+2KLoC) to make a reference backend faster, this is the wrong tradeoff for us

PR #725, opened 2026-08-17
Bound MoE routing to kernel capacity

xhjkl:
MoE routing uses fixed threadgroup storage for expert logits, TopK selection, and model input caching. This change centralizes those limits across Rust and Metal and rejects configurations that exceed them before parameter loading or dispatch.

It also validates nonzero expert dimensions, checked hidden-width expansion, and route-count arithmetic without changing the existing MoE execution pipeline.

### DIFF
--- a/crates/backend-uzu/build/metal/artifact_cache.rs
+++ b/crates/backend-uzu/build/metal/artifact_cache.rs
@@ -1,0 +1,49 @@
+use std::{collections::HashMap, fs};
+
+use anyhow::Context;
+use serde::{Deserialize, Serialize};
+
+pub use super::cache_protocol::{SharedArtifactCache, air_key, dependencies_match, index_key, metallib_key, zstd_key};
+use super::{ast::MetalKernelInfo, cache_protocol::Stage};
+use crate::common::{caching, identifiers::KernelName, kernel::Kernel};
+
+#[derive(Serialize, Deserialize, Clone)]
+pub struct SourceIndex {
+    pub schema: u32,
+    pub source_hash: [u8; blake3::OUT_LEN],
+    pub compile_schema_hash: [u8; blake3::OUT_LEN],
+    pub analyzer_hash: [u8; blake3::OUT_LEN],
+    pub gpu_types_hash: [u8; blake3::OUT_LEN],
+    pub dependency_hashes: HashMap<Box<str>, [u8; blake3::OUT_LEN]>,
+    pub footer: String,
+    pub kernel_infos: Vec<MetalKernelInfo>,
+    pub specialize_indices: HashMap<KernelName, usize>,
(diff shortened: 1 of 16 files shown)


crates/backend-uzu/build/metal/artifact_cache.rs
```diff
@@ -0,0 +1,49 @@
+use std::{collections::HashMap, fs};
```
CC-Yeh, line 1: Can you elaborate why do we need this exactly?

crates/backend-uzu/build/metal/cache_protocol.rs
```diff
@@ -0,0 +1,321 @@
+use std::{
```
CC-Yeh, line 1: Same, why do we need this? and this change should not be included in this PR.

crates/backend-uzu/tests/unit/encodable_block/moe/moe_block_validation_test.rs
```diff
@@ -0,0 +1,88 @@
+use std::io::Write;
```
CC-Yeh, line 1: Don't think we need all these new lines to check very simple logic

CC-Yeh (changes requested):
Can you please remove irrelevant changes?

CC-Yeh (commented):
The limits aren't centralized for Metal parts:

- `512` / `128` stay as separate literals in `moe/mod.rs:93,97` and
  `router_topk.metal:9,10`; the diff rewrites the line holding 512 and
  leaves it a literal.
- A const-only `gpu_types/moe.rs` auto-generates `generated/moe.h`
  (pattern: `hadamard_order.rs`, no build wiring)
- The third limit is unbounded: `router_topk.metal:35` declares
  `x_cache[1024]`, `:56` writes `x_cache[c]` for `c < d_model/4` with no
  clamp, and Rust only checks `model_dim % 4 == 0`. `model_dim > 4096`
  overruns a 16 KB threadgroup buffer from a config file.

xhjkl:
Superseded with #746.

crates/backend-uzu/build/metal/artifact_cache.rs
xhjkl: I'm so sorry: it's indeed an irrelevant change to the PR subject. My intention was to stack those PR together so that you could review or merge then in independently. So I tried to structure them as:
* `build cache` merges into `main`,
* then `mxfp4 groundwork` merges into `build cache`,
* then `proper mxfp4 impl` merges into `mxfp4 groundwork`, and so on the they chain one off another.

But Github apparently won't allow that because I'm not a repo owner of Uzu, so my next PR is a single flat submission, not a chain anymore. Hope this works 🙏

PR #726, opened 2026-08-17
Route expert rows directly through matmul

xhjkl:
Matmul can now consume token-major TopK expert IDs directly, selecting one expert matrix for each route row and producing route-major output. Dense matmul, sparse readout, and expert routing are represented as distinct routing modes, while per-expert bias remains an output operation rather than routing metadata.

The CPU reference and shared Metal GEMV path cover token-row reuse, route-major inputs, small output widths, invalid expert IDs, large expert banks, and checked buffer bounds. No MoE pipeline or microfloat format is introduced here.


xhjkl:
Superseded with #746.

PR #727, opened 2026-08-17
Execute MoE directly from TopK routes

xhjkl:
MoE execution is now expressed as `router.route(input)` followed by `experts.encode(input, routes)`. Gate/up reads token row `route / K`, down projection reads route rows directly, and route-weighted reduction consumes route-major expert output.

This removes counts, offsets, buckets, gathers, scatter maps, and separate decode/prefill expert implementations from the active path. Full-precision decode and prefill share the same expert interface, with an independent scalar oracle covering expert selection, gate/up ordering, biases, clipping, and weighted reduction.


xhjkl:
Superseded with #746.

PR #728, opened 2026-08-17
Load MXFP4 expert banks without unpacking

xhjkl:
Expert projections can now load MXFP4 matrix banks from their packed E2M1 values, E8M0 block scales, and per-matrix outer scales. The packed operands execute through the same direct expert-routing and `MoeExperts` control flow as full-precision weights, without a separate MoE kernel family or BF16 expansion.

The initial runtime supports the GPT-OSS layouts needed here—group-16 W13 and group-32 W2—and validates them against independent scalar decoding and artifact-shaped fixtures. Microfloat use outside expert banks and NVFP4 are rejected explicitly rather than implying unsupported coverage.


xhjkl:
Superseded with #746.

PR #740, opened 2026-08-18
Cache compiled Metal artifacts across builds

xhjkl:
This change set adds a target-wide content-addressed cache. Analysis, AIR, metallib, and compressed metallib outputs are keyed separately by their actual source, dependency, generated-code, toolchain, flag, and compression inputs. Distinct Cargo units can therefore reuse Metal artifacts whenever those inputs are unchanged.

In a release build, enabling the unrelated `grammar` feature improved the second build from ~13m to ~40s, with 63 hit, 0 miss.

### DIFF
--- a/crates/backend-uzu/Cargo.toml
+++ b/crates/backend-uzu/Cargo.toml
@@ -61,6 +61,7 @@ shoji.workspace = true
 half.workspace = true
 serde.workspace = true
 serde_json.workspace = true
+tokio = { workspace = true, features = ["macros", "rt", "time"] }
 monostate.workspace = true
 thiserror.workspace = true
 tokenizers.workspace = true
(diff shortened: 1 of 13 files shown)


uuuvn:
What does "equivalent Cargo build units" means exactly? How can a provably equivalent yet cache miss build happen?

crates/backend-uzu/Cargo.toml
```diff
@@ -61,6 +61,7 @@ shoji.workspace = true
 half.workspace = true
 serde.workspace = true
 serde_json.workspace = true
+tokio = { workspace = true, features = ["macros", "rt", "time"] }
```
uuuvn: Also why is this adding tokio to main crate dependencies? You're not touching anything inside src/ in this pr, only build/ and tests/

xhjkl:
It's not functioning as intended, yet it does add complexity. We had better close it unmerged.

PR #743, opened 2026-08-19
Compare UZU Mamtul vs MLX 

eugenebokhan:
```shell
cargo +nightly test -p backend-uzu --lib --features mlx-compare -- \
  --ignored --nocapture quantized_matmul_benchmark bits=4,8 g=32 k=2560 n=2560
```

Screenshot 2026-08-19 at 17 21 38

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -473,6 +473,7 @@ dependencies = [
  "blake3",
  "bytemuck",
  "clang-ast",
+ "comfy-table",
  "criterion",
  "depfile",
  "derive_more",
@@ -482,6 +483,8 @@ dependencies = [
  "itertools 0.14.0",
  "libm",
  "memmap2",
+ "mlx-rs",
+ "mlx-sys",
  "monostate",
  "mtl-rs",
  "ndarray",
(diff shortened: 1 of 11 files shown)


CC-Yeh (commented):
Nice one!

A few things to keep two columns from being comparable:

**Dispatch semantics:** uzu uses `MTLDispatchTypeSerial` afaik, and mlx uses `MTLDispatchTypeConcurrent`. Then 100 dispatches run for uzu is sequential and for mlx 100 independent matmuls can overlap freely.

**Output buffers:** uzu reuses one allocation while MLX allocates a fresh output per call, all 100 alive at once. Could impact larger ms

**Cache Residency** current weights re-read 100x and fits in SLC, kernels should read data from DRAM (cold) not SLC to mimic real decode scenario.

**Not kernel-to-kernel**: it should compare only matching quantization format speed, it's okay to list all of them in same table I think but need to flag what uzu formats map to mlx formats.

CC-Yeh:
with fix: Chain each dispatch onto the previous one's output in both engines, so MLX's 100 independent matmuls can't silently overlap while uzu's serial encoder runs them one at a time

```
cargo +nightly test --release -p backend-uzu --lib --features mlx-compare -- \
  --ignored --nocapture quantized_matmul_benchmark bits=4,8 g=32 k=2560 n=2560
```

On M4, this is pre new qmv, not sure if mlx contains batched qmv yet
Screenshot 2026-08-20 at 2 30 14 PM

Screenshot 2026-08-20 at 2 30 29 PM

PR #746, opened 2026-08-20
Enable GPT-OSS inference with native MXFP4 MoE

xhjkl:
This adds native MXFP4 expert banks to Uzu so it can load and run GPT-OSS and similarly packed MoE models without expanding their weights.

----

Notable changes:
```
// weight_matrix.rs
// before
weights = FullPrecision | IntegerQuantized;

// after
weights = FullPrecision | IntegerQuantized | MXFP4 {
    e2m1_codes,
    e8m0_scales,
    outer_scales,
};
// MXFP4 is a first-class matrix format, independent of whether
// the matrix is dense, sparsely read, or part of an expert bank.
```

```
// moe/mod.rs
// before
router
    -> counts
    -> offsets
    -> buckets
    -> permutation
    -> gathered_inputs
    -> experts;

// after
let routes = router.route(input, token_count, encoder)?;
let output = experts.encode(input, &routes, encoder)?;
// Top-K expert IDs are now the contract; the old public
// bookkeeping and materialized input expansion are gone.
```

```
// metal/matmul/mod.rs
match select_dispatch(shape) {
    RoutedMxfp4Decode => fused_mxfp4_gemv(...),
    SmallWorkload     => gathered_gemv(...),
    Prefill           => row_tiled_gathered_gemm(...),
}
// Decode and prefill meet at one matmul seam. The prefill arm
// reuses each selected expert tile across route rows, including MXFP4.
```

```
// attention + chat
let qkvg = fused_projection.encode(hidden, ...)?;
let qkvg = normalize_qk_and_apply_sigmoid_gate(qkvg)?;
let reply = harmony.decode(stream)?;
// GPT-OSS fused attention projections and Harmony framing complete
// the model path around the new expert execution.
```


xhjkl:
Decided to split it down to serial chunks, so not merging this one in.

The next ones are:
* #699,
* #747.

PR #747, opened 2026-08-20
Honor configured SiLU alpha and clipping in GatedActMul

xhjkl:
This makes the existing gated-MLP activation settings effective. Dense MLPs now apply non-default SiLU alpha, gate clipping, and up/value clipping instead of silently using default SiLU with unbounded operands.

By keeping activation semantics at the shared `GatedActMul` seam, this also lays the groundwork for MXFP4 expert banks without coupling their weight format to MLP post-processing.

Notable path:

```text
DenseMLPConfig
  ├─ activation.alpha
  ├─ gate_clipping
  └─ up_clipping
       ↓
MlpGateActMulEncodable
       ↓
GatedActMulSettings
       ↓
full-precision or activation-quantized GatedActMul
       ↓
CPU / Metal specialization
```

Clipping is applied before gate activation and value multiplication. Keeping these transformations in one settings object ensures full-precision and activation-quantized paths use identical semantics. Per-layer embeddings now also honor non-default SiLU alpha.

Tests cover transformed f32 and bf16 execution across CPU and Metal while retaining coverage for the default and Hadamard paths.


crates/backend-uzu/tests/unit/backends/common/kernel/gated_act_mul_test.rs
```diff
@@ -97,7 +102,9 @@ fn test_gated_act_mul_interleaved_hadamard_bf16() {
     let expected = run_interleaved::<bf16, Cpu>(&input, true);
     for_each_non_cpu_backend!(|B| {
         let actual = run_interleaved::<bf16, B>(&input, true);
-        assert_eq_float::<bf16>(&expected, &actual, 0.02, "Hadamard gated activation mismatch");
+        // CPU and Metal can round the bf16 product at different points before
+        // the f32 RHT, leaving valid results one bf16 ULP apart here.
+        assert_eq_float::<bf16>(&expected, &actual, 0.04, "Hadamard gated activation mismatch");
```
CC-Yeh: This assertion should pass before your changes. So I think sth changes elsewhere cause this.

crates/backend-uzu/src/backends/metal/kernel/gated_act_mul/gated_act_mul.metal
```diff
@@ -60,11 +68,20 @@ PUBLIC KERNEL(GatedActMul) (
       gate = act_operand[batch_idx * gated_dim + gated_idx];
     }
   }
+  if (clip_gate) {
+    gate = static_cast<T>(clamp(float(gate), gate_clip_min, gate_clip_max));
+  }
+  if (clip_value) {
+    value = static_cast<T>(clamp(float(value), value_clip_min, value_clip_max));
+  }
```
CC-Yeh, line 76: Why cast it back to T?

crates/uzu-engine/src/backends/common/kernel/gated_act_mul.rs
```diff
@@ -33,13 +33,28 @@ bitflags! {
...
+/// Value transforms baked into a gated-activation kernel specialization.
+#[derive(Debug, Clone, Copy, Default)]
+pub struct GatedActMulSettings {
+    /// Non-default slope for SiLU's sigmoid term.
+    pub activation_alpha: Option<f32>,
+    /// Bounds applied to the gate before activation.
+    pub gate_clipping: Option<(f32, f32)>,
+    /// Bounds applied to the value before multiplication.
+    pub value_clipping: Option<(f32, f32)>,
+}
```
CC-Yeh, line 46: Variables are clear enough, comments looks redundant

crates/backend-uzu/src/backends/common/kernel/gated_act_mul.rs
```diff
@@ -98,11 +122,15 @@ impl<B: Backend> GatedActMul<B> {
             options.contains(GatedActMulOptions::HADAMARD),
             activation_group_size,
             sum_group_size,
+            options.contains(GatedActMulOptions::CUSTOM_ALPHA),
+            options.contains(GatedActMulOptions::CLIP_GATE),
+            options.contains(GatedActMulOptions::CLIP_VALUE),
```
CC-Yeh: Lines 114-116 set three bitflag bits from settings.*.is_some(), then lines 122-124 read those same bits straight back — and the bits are stored in self.options where nothing ever reads them again. Pass the predicates directly and delete the three constants.

crates/backend-uzu/src/backends/common/kernel/gated_act_mul.rs
```diff
@@ -141,6 +169,11 @@ impl<B: Backend> GatedActMul<B> {
             value_offset,
             value_row_stride,
             act_type,
+            self.settings.activation_alpha,
+            self.settings.gate_clipping.map(|(min, _)| min),
+            self.settings.gate_clipping.map(|(_, max)| max),
+            self.settings.value_clipping.map(|(min, _)| min),
+            self.settings.value_clipping.map(|(_, max)| max),
```
CC-Yeh: A bit hard to read, maybe a unzip helper is better

crates/backend-uzu/src/encodable_block/mlp/gate_act_mul.rs
```diff
@@ -116,4 +126,13 @@ impl<B: Backend> MlpGateActMulEncodable<B> {
...
         Ok(input)
     }
+
+    fn clipping_bounds(clipping: Option<(Option<f32>, Option<f32>)>) -> Option<(f32, f32)> {
+        let (min, max) = clipping?;
+        if min.is_none() && max.is_none() {
+            return None;
+        }
+
+        Some((min.unwrap_or(f32::NEG_INFINITY), max.unwrap_or(f32::INFINITY)))
```
CC-Yeh: How does metal handle inf/-inf with clamp? with and without -fast-math?
Do you have test for this?

CC-Yeh (changes requested):

CC-Yeh:
@codex review

crates/backend-uzu/tests/unit/backends/common/kernel/gated_act_mul_test.rs
xhjkl: Indeed, on my machine the discrepancy predates these changes:
```
$ git checkout 9c405b16e # ← upstream main before the fix
$ cargo test -p backend-uzu test_gated_act_mul_interleaved_hadamard_bf16
[...]
thread 'backend_uzu::backends::common::kernel::tests::gated_act_mul_test::test_gated_act_mul_interleaved_hadamard_bf16' (10108443) panicked at crates/backend-uzu/src/tests/assert.rs:27:9:
Hadamard gated activation mismatch. Mismatch at index 24: expected -6.4375, got -6.46875, diff 0.03125 (eps 0.02)
[...]
```

So let's keep it as it is, perhaps, and let the CI be our judge. But otherwise if you want me to keep it at `0.04`, please let me know.

crates/uzu-engine/src/backends/common/kernel/gated_act_mul.rs
xhjkl: Agreed. Trimmed down the commentary now.

crates/backend-uzu/src/backends/common/kernel/gated_act_mul.rs
xhjkl: Eeek, my bad! Good catch; now fixed.

crates/backend-uzu/src/backends/common/kernel/gated_act_mul.rs
xhjkl: Agreed. Settled on destructuring the bounds instead.

crates/backend-uzu/src/encodable_block/mlp/gate_act_mul.rs
xhjkl: Just [checked empirically on my machine](https://gist.github.com/xhjkl/89b4b22f1f6ac5697c43518e0d655cbf). All the compile modes (O2, fast-math, no-fast-math, relaxed mode, safe mode) do agree:
* `clamp(+inf, -inf, +inf)` → `+inf` (`0x7f800000`)
* `clamp(-inf, -inf, +inf)` → `-inf` (`0xff800000`)
* `clamp(+inf, -inf, 2)` → `2.0` (`0x40000000`)
* `clamp(-inf, -2, +inf)` → `-2.0` (`0xc0000000`)

However, the revised implementation no longer relies on sentinels and applies the bounds independently.

crates/backend-uzu/src/backends/metal/kernel/gated_act_mul/gated_act_mul.metal
xhjkl: I cast back to `T` to preserve the kernel's existing dtype. That `float` is temporary and is only needed for applying the `float` clipping bounds. llama-cpp similarly widens and clips per-element inside its [metal kernel](https://github.com/ggml-org/llama.cpp/blob/17197474510622a3b4ea7d0909d70b606f542b96/ggml/src/ggml-metal/ggml-metal.metal#L1627-L1656), although it then keeps the activation and multiplication in `float` and narrows only the final result. Our earlier narrowing is deliberate, even if not strictly required by `gpt-oss`. I'm open to following llama-cpp's choice if you prefer it that way.

crates/backend-uzu/src/encodable_block/mlp/gate_act_mul.rs
```diff
@@ -28,14 +29,24 @@ impl<B: Backend> MlpGateActMulEncodable<B> {
...
+        value_clipping: ClippingBounds,
         hidden_dim: u32,
         input_preparation: Option<LinearInputPreparation<B>>,
     ) -> Result<Self, B::Error> {
         let (hadamard_factors, a8_plan) = input_preparation
             .map_or((None, None), |preparation| (Some(preparation.input_factors), preparation.a8_plan));
-        let fp_kernel = GatedActMul::full_precision(context, data_type, true, hadamard_factors.is_some())?;
+        let activation_alpha = activation.alpha();
+        let settings = GatedActMulSettings {
+            activation_alpha: (activation_alpha != 1.0).then_some(activation_alpha),
```
CC-Yeh: I think this logic is worth moving somewhere closer to activation config

sth like:
```
pub fn custom_alpha(&self) -> Option<f32> {
    match self {
        AnyActivation::SiLU(config) if config.alpha != 1.0 => Some(config.alpha),
        _ => None,
    }
}
```

crates/backend-uzu/src/clipping.rs
```diff
@@ -0,0 +1,73 @@
...
+
+    /// Clips a value to the configured bounds.
+    pub fn apply(
+        self,
+        value: f32,
+    ) -> f32 {
+        let value = self.min.map_or(value, |min| value.max(min));
+        self.max.map_or(value, |max| value.min(max))
+    }
+}
```
CC-Yeh: I don't think all these functions have consumers besides test, let's add them back when really needed

crates/backend-uzu/src/lib.rs
```diff
@@ -2,6 +2,7 @@
 #![cfg_attr(test, test_runner(test_runner::uzu_harness))]
 
 mod array;
+mod clipping;
```
CC-Yeh: I think clipping should move under config?

crates/backend-uzu/src/backends/metal/kernel/gated_act_mul/gated_act_mul.metal
```diff
@@ -31,11 +31,21 @@ PUBLIC KERNEL(GatedActMul) (
     const constant uint& value_offset,
     const constant uint& value_row_stride,
     const constant ActivationType& act_type,
+    const constant float& activation_alpha OPTIONAL(custom_activation_alpha),
+    const constant float& gate_clip_min OPTIONAL(clip_gate_min),
+    const constant float& gate_clip_max OPTIONAL(clip_gate_max),
+    const constant float& value_clip_min OPTIONAL(clip_value_min),
+    const constant float& value_clip_max OPTIONAL(clip_value_max),
```
CC-Yeh: I like the old way better (no _min/_max), too many function constants may blow up PSO cache.

crates/backend-uzu/src/backends/metal/kernel/gated_act_mul/gated_act_mul.metal
```diff
@@ -60,19 +70,45 @@ PUBLIC KERNEL(GatedActMul) (
...
+  if (clip_value_min || clip_value_max) {
+    float clipped_value = float(value);
+    if (clip_value_min) {
+      clipped_value = max(clipped_value, value_clip_min);
+    }
+    if (clip_value_max) {
+      clipped_value = min(clipped_value, value_clip_max);
+    }
+    value = static_cast<T>(clipped_value);
+  }
```
CC-Yeh: Let's just use `clamp(val, min, max)`, and clipping can set default min = f32::MIN and f32::MAX (for bf16 as well, when construct clipping we already know what data type is T i think), so no potential undefined issue. would be cleaner, and clamp and current both lower to minmax ig

CC-Yeh (changes requested):

crates/backend-uzu/src/backends/metal/kernel/gated_act_mul/gated_act_mul.metal
xhjkl: Understood. Redone it that way.

CC-Yeh (approved):
LGTM, thanks!

CC-Yeh:
@xhjkl can you fix the linter errors thanks?

xhjkl:
@CC-Yeh, all done ✨

PR #748, opened 2026-08-20
Prepare for the big high level crate rewrite

uuuvn:


eugenebokhan:
are you up to rewrite all from scratch? let's do incrementally

uuuvn:
> are you up to rewrite all from scratch? let's do incrementally

It'll be from scratch but incrementally. New crates are going to be built one by one then just switched and legacy/ rm -rfd

PR #752, opened 2026-08-22
fix grammar leak across streams

gearonixx:
This PR tries to fix:

```rust
  // TODO: this leaks previous LanguageModelStreamOptions
  ```
  
The current issue is that a pending `output token` in `stream.rs` can cross a stream boundary even when the new stream uses a different grammar.

```text
Stream #1
grammar = text / None
→ pending token = "history"

Stream #2
grammar = JSON
→ inherits "history" - very first token is printed with the Stream #1 grammar 
→ then continues under JSON grammar

Result:
history{"name":"Bob"} - WRONG
```

A stream with a grammar should only return output that matches that grammar.

`history{"name":"Bob"}` is invalid JSON because the first token was sampled using the **previous stream**'s grammar config. (the first token was cooked using the old grammar)

Now: 

```text
same grammar config
→ replay committed tokens into the new Grammar matcher
→ continue

input is empty
→ pending output exists?
   compare grammars

or maybe new stream has no grammar
→ continue normally
```

Sampling method differences are intentionally not treated as incompatible here. this PR only addresses grammar correctness.

### DIFF
--- a/crates/backend-uzu/src/engine/language_model/stream/stream.rs
+++ b/crates/backend-uzu/src/engine/language_model/stream/stream.rs
@@ -20,7 +20,7 @@ use crate::{
         capture::CaptureSpan,
         language_model::{
             LanguageModel,
-            state::LanguageModelState,
+            state::{LanguageModelState, PendingOutput},
             stream::{LanguageModelStreamError, LanguageModelStreamOptions},
         },
     },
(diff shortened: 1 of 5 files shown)


crates/backend-uzu/src/engine/language_model/stream/stream.rs
```diff
@@ -343,9 +347,20 @@ impl<'a, B: Backend> LanguageModelStream<'a, B> {
                 output_tokens: output_tokens.unwrap(),
             })
         } else {
-            // TODO: this leaks previous LanguageModelStreamOptions
+            let pending = model_state.last_output.take().unwrap();
+            #[cfg(grammar)]
+            match (&pending.grammar, options.grammar.as_ref().map(Grammar::config)) {
+                (old, Some(new)) if old.as_ref() == Some(new) => {
+                    let grammar = options.grammar.as_mut().unwrap();
+                    for &token in &model_state.tokens[model_state.grammar_start..] {
```
gearonixx, line 355: the grammar object is new, so its matcher knows nothing about what was already generated. Replaying `tokens[grammar_start..]` brings it up to date

uuuvn:
If you keep sending nonsensical prs you will get banned.

PR #753, opened 2026-08-22
Wire sampling seed

CC-Yeh:
good for deterministic experiments

### DIFF
--- a/crates/backend-uzu/src/engine/language_model/state.rs
+++ b/crates/backend-uzu/src/engine/language_model/state.rs
@@ -31,19 +31,12 @@ impl<B: Backend> LanguageModel<B> {
     pub fn create_empty_state(
         &self,
         max_context_length: Option<u32>,
-    ) -> Result<LanguageModelState<B>, LanguageModelCreateEmptyStateError<B>> {
-        self.create_empty_state_with_seed(max_context_length, None)
-    }
-
-    pub(crate) fn create_empty_state_with_seed(
-        &self,
-        max_context_length: Option<u32>,
-        sampling_seed: Option<u64>,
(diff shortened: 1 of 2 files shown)


CC-Yeh:
@codex review

crates/backend-uzu/src/engine/language_model/state.rs
```diff
@@ -31,11 +31,19 @@ impl<B: Backend> LanguageModel<B> {
...
         max_context_length: Option<u32>,
+    ) -> Result<LanguageModelState<B>, LanguageModelCreateEmptyStateError<B>> {
+        self.create_empty_state_with_seed(max_context_length, None)
+    }
+
+    pub(crate) fn create_empty_state_with_seed(
+        &self,
+        max_context_length: Option<u32>,
+        sampling_seed: Option<u64>,
     ) -> Result<LanguageModelState<B>, LanguageModelCreateEmptyStateError<B>> {
```
uuuvn: Why is this `(crate)`? Also why not just add `sampling_seed: u64` to existing create_empty_state and move the rand::random to SamplingSeed::Default match arm?

uuuvn (changes requested):

crates/backend-uzu/src/engine/language_model/state.rs
uuuvn: As is this pr also leaves/adds create_empty_state as dead code... This sounds like an llm compatibility slop, please be more careful

CC-Yeh: Fixed, yeah should've taken more time, my fault

uuuvn (approved):

PR #761, opened 2026-08-24
Support ring KV caches for noncausal attention

CC-Yeh:


uuuvn:
There is no non-causal sliding window in our dflash

CC-Yeh:
> There is no non-causal sliding window in our dflash

Yes, was using lalamo pre your fix.

but still worth keeping for any noncausal sliding window attention, wdyt? @uuuvn

uuuvn:
> > There is no non-causal sliding window in our dflash
> 
> Yes, was using lalamo pre your fix.
> 
> but still worth keeping for any noncausal sliding window attention, wdyt? @uuuvn

If it doesn't introduce a ton of complexity (or even better makes things simpler too) then yeah. Will take a look at the code soon-ish

uuuvn:
I don't recall why I didn't do ring for non-causal, I think I asserted that swa is always causal in the big refactor attention port, but I think it makes much more sense for it to be ring too

crates/uzu-engine/src/encodable_block/mixer/attention/mod.rs
```diff
@@ -164,7 +163,8 @@ impl<B: Backend> Attention<B> {
             .then(|| parameter_tree.leaf("sinks")?.validate(&[num_q_heads], data_type)?.read_allocation())
             .transpose()?;
 
-        let is_kv_cache_ring = is_causal && sliding_window_size.is_some();
+        let ring_capacity = sliding_window_size.filter(|capacity| *capacity > 0);
```
uuuvn: `sliding_window_size = Some(0)` is nonsense that definitely shouldn't just be silently interpreted as  `None` / no sliding window. Either ignore assuming callers won't pass it or just assert (already done in KVCacheState::ring) for hardening.

I'm not sure if this one is llm or not but they love to add dangerous silent clamps of nonsensical arguments into something else and happily proceeding instead of failing loudly (which I assume is unfortunately quite advantageous in rlvr...), please be more careful.

uuuvn (changes requested):

crates/uzu-engine/src/encodable_block/mixer/attention/mod.rs
CC-Yeh: Nice catch thx. Added an assertion

PR #765, opened 2026-08-24
Add AttentionGemmGrouped

CC-Yeh:
GQA-fused attention kernel for long-context specdec

- One kernel: decode (S≤64) through prefill (S≤1024)
- Causal, non-causal, and trie masks (trie: S≤64)
- GQA heads share one KV read per group
- Split-KV with numerically stable softmax merge fills all cores
- MXU at head_dim 256 via register-safe 2D tiling
- Split counts scale with GPU core count
- Measured per-cell routing; non-MXU chips unchanged

Cleanups:

- Simpler dispatch
- Only single pass kernel for CPU
- Remove unused half type


CC-Yeh:
@codex review

crates/uzu-engine/src/backends/metal/kernel/attention/gemm_grouped/attention_gemm_grouped.metal
```diff
@@ -0,0 +1,452 @@
...
+        }
+        return visible ? value : masked_score_pre_scale;
+      });
+    }
+
+    AccumType block_max[ROWS_PER_LANE];
+    score.row_reduce(block_max, -INFINITY, [](AccumType a, AccumType b) { return metal::max(a, b); });
+    METAL_PRAGMA_UNROLL
+    for (int r = 0; r < ROWS_PER_LANE; ++r) {
+      block_max[r] *= score_scale;
```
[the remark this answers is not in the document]
CC-Yeh: in production scale should be positive, guarded

CC-Yeh:
@codex review

uuuvn:
Dispatching logic is getting complicated again, let's move the whole many cores thing into the backends. So there is one manual attention "kernel" like there is one manual matmul "kernel" that does all the dispatching inside per-backend. Cpu always does the simplest naivest attention in it, metal dispatches into one of it's (now private) kernels

CC-Yeh:
@uuuvn cleaned up the attention routing and removed CPU attention kernels except SinglePass

crates/uzu-engine/src/backends/common/kernel/attention.rs
```diff
@@ -0,0 +1,44 @@
+use crate::{
+    backends::common::{Allocation, Backend, BufferArg, Encoder},
+    data_type::DataType,
+    encodable_block::mixer::attention::AttentionStateType,
+};
+
+pub trait AttentionKernel<B: Backend>: Sized + Send + Sync {
```
uuuvn: This is not correct, attention kernel is not generic, it should have an associated type backend with two-way binding. See how it's done in other kernels like MatmulKernel.

uuuvn (changes requested):

crates/uzu-engine/src/backends/common/kernel/attention.rs
uuuvn: Actually all the manual kernels other than `type MatmulKernel: matmul::MatmulKernel<Backend = Self::Backend>;` are done wrongly too, this is one of the things to do redo properly when cleaning things up

uuuvn: Oh the other ones are generic because they're optional and set `Infallible`, but non-optional ones should definitely be associated types and optional ones probably should just have a `NoKernel<B>` helper type

CC-Yeh: Thanks for the catch, associated with Backend

PR #767, opened 2026-08-25
Bound tool-call test generations

xhjkl:
Tool-call integration tests currently leave their per-turn generation budget unset. A malformed response that never emits its stop token can therefore decode toward the model context limit and make cargo test appear hung.

This caps each generation at 256 tokens. Correct short tool conversations are unchanged; missing-stop regressions now return Length and fail the existing Stop assertion promptly.

Validation:
- cargo +nightly fmt --check
- cargo +nightly clippy -p uzu --test tool -- -D warnings
- On the reproducing host, calls::qwen3_5_0_8b now reaches the new limit and reports the offending third prompt in 10.96 seconds instead of continuing indefinitely.


xhjkl:
That's an op issue on my end, the mainline is probably fine without the token limit.

PR #768, opened 2026-08-25
Enforce fused TopK router capacities

xhjkl:
This lifts the fused Metal TopK router’s existing implicit capacities into one place and enforces them on the host before dispatch. Valid configurations behave exactly as before; configurations that would exceed the kernel’s statically sized threadgroup buffers now fail validation instead of scheduling an unsafe compute call.

Notable changes:
- the GPU-type build step generates router_topk.h from the Rust capacity constants
- MoeBlock validation bounds model width, routed expert count, and active TopK width against the kernel’s statically sized buffers
- focused unit tests cover accepted boundary values and rejected overflow configurations


crates/uzu-engine/src/backends/common/gpu_types/router_topk.rs
```diff
@@ -0,0 +1,11 @@
+/// Maximum expert logits retained by the fused router/TopK kernel.
+pub const ROUTER_TOPK_MAX_EXPERTS: u32 = 512;
+
+/// Maximum selection width supported by the fused TopK router kernel.
+pub const ROUTER_TOPK_MAX_SELECTED_EXPERTS: u32 = 128;
+
+/// Maximum model width cached by the fused TopK router kernel.
+pub const ROUTER_TOPK_MAX_MODEL_DIM: u32 = 4096;
```
uuuvn: Those comments are way too verbose and unnecessary, all const names are self-explanatory here

crates/uzu-engine/src/backends/common/gpu_types/router_topk.rs
```diff
@@ -0,0 +1,11 @@
...
+
+/// Threads launched by one fused TopK router threadgroup.
+pub const ROUTER_TOPK_THREADS_PER_THREADGROUP: u32 = 256;
```
uuuvn: This is a metal implementation detail that shouldn't leak into anything outside of metal backend

crates/uzu-engine/src/backends/metal/kernel/moe/router_topk.metal
```diff
@@ -32,15 +31,16 @@ PUBLIC KERNEL(MoeRouterTopK)(
...
+    threadgroup float4 x_cache[ROUTER_TOPK_MAX_MODEL_DIM / 4],
+    threadgroup float logits_shared[ROUTER_TOPK_MAX_EXPERTS],
+    threadgroup float reduce_tmp[ROUTER_TOPK_THREADS_PER_THREADGROUP],
+    threadgroup uint reduce_tmp_u[ROUTER_TOPK_THREADS_PER_THREADGROUP],
     threadgroup uint shared_best_idx[1],
     threadgroup float shared_best_val[1],
     const ThreadContext thread_context,
     const uint tgpig_x GROUPS(1),
     const uint token_idx GROUPS(t),
+    // Host dispatch expressions cannot reference shader constants; keep this literal aligned with gpu_types/router_topk.
```
uuuvn: Verbose llm comment. This one is worth keeping (preferably on the same line as THREADS) but should be shorter

crates/uzu-engine/src/encodable_block/mlp/moe/mod.rs
```diff
@@ -79,6 +82,24 @@ pub enum MoeBlockError<B: Backend> {
...
 }
 
+pub(crate) fn valid_model_dim(model_dim: u32) -> bool {
+    model_dim > 0 && model_dim <= ROUTER_TOPK_MAX_MODEL_DIM && model_dim.is_multiple_of(4)
+}
+
+pub(crate) fn validate_expert_counts<B: Backend>(
+    routed_experts: u32,
+    active_experts: u32,
+) -> Result<(), MoeBlockError<B>> {
```
uuuvn: What's the meaning behind those `(crate)`? LLMs like to add meaningless things like that everywhere, please be more careful when using them.

crates/uzu-engine/src/encodable_block/mlp/moe/mod.rs
```diff
@@ -79,6 +82,24 @@ pub enum MoeBlockError<B: Backend> {
...
+    if !(1..=ROUTER_TOPK_MAX_EXPERTS).contains(&routed_experts) {
+        return Err(MoeBlockError::InvalidRoutedExpertCount);
+    }
+    if !(1..=ROUTER_TOPK_MAX_SELECTED_EXPERTS.min(routed_experts)).contains(&active_experts) {
+        return Err(MoeBlockError::InvalidActiveExpertCount);
+    }
+
+    Ok(())
+}
```
uuuvn: These two validation helpers are used only once each in outside of tests and tests shouldn't use them

crates/uzu-engine/unit/encodable_block/moe/moe_block_validation_test.rs
```diff
@@ -0,0 +1,29 @@
...
+    }
+}
+
+#[uzu_test]
+fn rejects_router_dimensions_beyond_its_threadgroup_cache() {
+    assert!(valid_model_dim(4096));
+    for model_dim in [0, 2, 4100] {
+        assert!(!valid_model_dim(model_dim));
+    }
+}
```
uuuvn: These tests are not very useful, you're just hardcoding a couple trivial examples of behavior of the function. I can't think of any realistic scenario how they would fail and catch a real bug

uuuvn (changes requested):

crates/uzu-engine/src/encodable_block/mlp/moe/mod.rs
xhjkl: Indeed, those should have been module-private, no `pub(crate)`. I'm sorry it slipped by me.

crates/uzu-engine/src/backends/common/gpu_types/router_topk.rs
xhjkl: Understood. Adding comments was actually something I did willingly — I irrationally like every definition commented, even if it's a medium-quality comment. But let's drop those altogether if you prefer it that way, I'm assuming the quality bar for existence of the commentary is to be non-obvious from the naming itself.

crates/uzu-engine/unit/encodable_block/moe/moe_block_validation_test.rs
xhjkl: Good point. So let's inline the validation methods into the constructor itself, and drop the trivial tests.

uuuvn (approved):

crates/uzu-engine/src/backends/metal/kernel/moe/router_topk.metal
```diff
@@ -32,16 +32,16 @@ PUBLIC KERNEL(MoeRouterTopK)(
...
+    threadgroup float logits_shared[ROUTER_TOPK_MAX_EXPERTS],
     threadgroup float reduce_tmp[THREADS_PER_TG],
     threadgroup uint reduce_tmp_u[THREADS_PER_TG],
     threadgroup uint shared_best_idx[1],
     threadgroup float shared_best_val[1],
     const ThreadContext thread_context,
     const uint tgpig_x GROUPS(1),
     const uint token_idx GROUPS(t),
-    const uint lid THREADS(256)
+    const uint lid THREADS(256) // Must match THREADS_PER_TG.
```
eugenebokhan: can't we simple declare a shared rust - metal constant and reuse if here?

eugenebokhan (approved):

crates/uzu-engine/src/backends/metal/kernel/moe/router_topk.metal
xhjkl: Good point. Replaced now the duplicated literal with a single Metal-local definition reused by `THREADS(...)`. This also keeps the details entirely inside the Metal part, reconciling it with [the earlier review item](https://github.com/trymirai/uzu/pull/768#discussion_r3857267564).

PR #769, opened 2026-08-25
Represent MXFP4 as a first-class weight format

xhjkl:
This allows Uzu to load, validate, and execute dense MXFP4 weight matrices through the CPU reference backend.

Metal execution follows in a separate slice, with optimized GEMV and GEMM kernels consuming the same MXFP4 operand representation. MXFP4 expert banks can then build on those paths.


crates/uzu-engine/src/backends/common/microfloat.rs
```diff
@@ -0,0 +1,227 @@
+use thiserror::Error;
```
eugenebokhan: let's make a mod called microfloat and inside have each file per type

crates/uzu-engine/src/backends/common/microfloat.rs
```diff
@@ -0,0 +1,227 @@
...
+#[inline]
+pub fn decode_mxfp4(
+    code: u8,
+    exponent: u8,
+    outer_scale: f32,
+) -> f32 {
+    decode_e2m1(code) * decode_e8m0(exponent) * outer_scale
+}
+
+#[cfg(test)]
```
eugenebokhan: unit test should be located in a dedicated dir and imported as file here

crates/uzu-engine/src/config/weight_matrix/microfloat_spec.rs
```diff
@@ -0,0 +1,36 @@
...
+
+#[uzu_config(super::WeightMatrixSpec)]
+pub struct MicrofloatSpec {
+    pub bits: u32,
+    pub group_size: usize,
+    pub scale_mode: MicrofloatFormat,
+    pub layout: Layout,
+}
+
+#[cfg(test)]
```
eugenebokhan: same here, import tests as file

crates/uzu-engine/src/config/weight_matrix/microfloat_spec.rs
```diff
@@ -0,0 +1,36 @@
...
+            "scale_mode": "mxfp4",
+            "layout": "output_input"
+        }))
+        .unwrap();
+        let AnyWeightMatrixSpec::MicrofloatSpec(spec) = spec else {
+            panic!("expected MicrofloatSpec");
+        };
+        assert_eq!(spec.scale_mode, MicrofloatFormat::Mxfp4);
+    }
+}
```
eugenebokhan: seems like unnecessary test

crates/uzu-engine/src/backends/common/microfloat.rs
```diff
@@ -0,0 +1,227 @@
...
+        self.encoding.bits()
+    }
+
+    pub fn group_size(self) -> u32 {
+        self.encoding.group_size()
+    }
+
+    pub fn rows(self) -> u32 {
+        self.rows
+    }
```
eugenebokhan: why not use `pub` fields?

crates/uzu-engine/src/encodable_block/weight_matrix.rs
```diff
@@ -29,13 +30,14 @@ pub struct QuantizationInfo {
 pub struct ParsedWeightSpec {
     pub layout: Layout,
     pub quantization: Option<QuantizationInfo>,
+    pub microfloat: Option<MicrofloatEncoding>,
```
eugenebokhan: seems like we need to extend the `QuantizationInfo` instead of sticking one more label to the spec

crates/uzu-engine/src/encodable_block/weight_matrix.rs
```diff
@@ -251,3 +329,96 @@ fn physical_shape(
         Layout::InputOutput => (input_dim, output_dim),
     }
 }
+
+#[cfg(test)]
```
eugenebokhan, line 347: import as file and think deeply if you really need all of them

eugenebokhan (changes requested):

crates/uzu-engine/src/backends/common/microfloat.rs
xhjkl: All done ✨

crates/uzu-engine/src/config/weight_matrix/microfloat_spec.rs
xhjkl: Agreed. Trimmed down the test only to those which cross domains.

crates/uzu-engine/src/backends/common/microfloat.rs
xhjkl: As you wish. Removed the accessors, made the fields public.

crates/uzu-engine/src/encodable_block/weight_matrix.rs
xhjkl: Totally agreed, that makes perfect sense. I'm so sorry I got vision-blurred on this one 🙏

eugenebokhan (approved):

PR #780, opened 2026-08-28
Update Device Profile

eugenebokhan:
Fotfix for running on iPhones


uuuvn (approved):
Uzu should still work even if it doesn't recognize the device, but fine as a hotfix to run on phones

PR #800, opened 2026-09-04
qtip s metal path

ry2009:
race kernels, structured tables, l15 masks, i3 hea…d sparse readout, tiered head loader, spec decode bench fields


ry2009:
just for cc for qtip metal kernels

ry2009:
Superseded by #864, which carries the S support (computed-codebook trellis kernels, restart-64 V4 and whole-row V2 leaves) onto current main in one PR.

PR #802, opened 2026-09-04
Fix tool calls result block parsing

### LINKED ISSUE #795: OpenAI-compatible server JSON-parses tool message content strings, crashing Qwen chat template with "Unexpected item type in content"
## Summary

The OpenAI-compatible server (`mirai server`) JSON-parses tool message `content` strings into native arrays before applying the model's chat template. When a tool result `content` string is a valid JSON array of objects, the Qwen 3.6 chat template receives an actual array instead of a string, iterates it as content blocks, and raises:

```
Render failed: invalid operation: Unexpected item type in content. (in chat_template:35)
```

The OpenAI Chat Completions API spec defines tool message `content` as a **string**. The Qwen chat template's `render_content` macro expects `content` to be either a string or an array of content blocks (`{type: "text", text: "..."}`, `{type: "image", ...}`). By parsing the string into a raw JSON array, uzu hands the template array items it cannot handle.

## Reproduction

Server: `mirai server --model trymirai/Qwen3.6-27B-M --port 8100`

Minimal failing request — a tool result whose `content` is a JSON-array string:

```bash
curl -s http://127.0.0.1:8100/v1/chat/completions \
  -H "Content-Type: application/json" \
  -H "Authorization: Bearer local" \
  -d '{
    "model": "alibaba:qwen3.6:27b:mirai:mirai-m:4",
    "max_tokens": 20,
    "messages": [
      {"role": "user", "content": "List issues"},
      {"role": "assistant", "content": "Fetch.", "reasoning_content": "thinking", "tool_calls": [
        {"id": "c1", "type": "function", "function": {"name": "bash", "arguments": "{}"}}
      ]},
      {"role": "tool", "content": "[{\"number\":228}]", "tool_call_id": "c1"},
      {"role": "user", "content": "ok"}
    ]
  }'
```

Response:

```json
{"error":{"message":"Render failed: invalid operation: Unexpected item type in content. (in chat_template:35)","type":"invalid_request_error","param":null,"code":"backend_error"}}
```

## What works vs. what fails

| Tool result `content` string | Result | Why |
|---|---|---|
| `"26"` | OK | Not valid JSON |
| `"{\"number\":1}"` | OK | JSON object, not array |
| `"[hello]"` | OK | Not valid JSON |
| `"[1,2,3]"` | **ERROR** | Parsed to `[1,2,3]`; integers fail the template's containment check |
| `"[{\"number\":228}]"` | **ERROR** | Parsed to array of objects; objects lack `text`/`image`/`video` keys |
| `"[{\"text\":\"hi\"}]"` | OK | Parsed to array; objects happen to have a `text` key the template accepts |

The pattern: uzu only parses strings that are valid JSON **arrays**. Objects, scalars, and non-JSON strings pass through correctly.

## Expected behavior

Tool message `content` should remain a string and be passed to the chat template as-is. The Qwen template handles string content directly:

```jinja
{%- if content is string %}
    {{- content }}
{%- elif content is iterable and content is not mapping %}
    {%- for item in content %}
        ...
```

## Actual behavior

uzu deserializes JSON-array-looking `content` strings into native arrays, so the template takes the `content is iterable` branch and iterates raw JSON objects as content blocks.

## Environment

- `mirai` CLI: `/Applications/Mirai.app/Contents/Resources/cli/mirai` (Mach-O arm64)
- Model: `trymirai/Qwen3.6-27B-M`
- Server: `mirai server --model trymirai/Qwen3.6-27B-M --port 8100`
- macOS (Apple Silicon)

## Impact

Any agent harness or client that returns JSON-array tool results (very common — e.g. `gh issue list --json`, any API returning lists) will crash the server once that tool result appears in conversation history. The failure is silent until the second turn that includes the result in context.

agolokoz:


uuuvn (approved):

kesslerio:
Confirmed fixed. Built from `main` (ce5881e9), started `mirai server --model trymirai/Qwen3.6-27B-M --port 8100`, and ran the full reproduction matrix from issue #795:

| Tool result `content` | Before | After |
|---|---|---|
| `"[{\"number\":228}]"` | crash | OK |
| `"[1,2,3]"` | crash | OK |
| `"[[{\"text\":\"hi\"}]]"` | crash | OK |
| `"26"`, `"{\"a\":1}"`, `"[hello]"`, `"true"`, `"null"`, `"42"` | OK | OK |

Also verified end-to-end through an agent harness (DeepSeek Harness): `gh issue list --json number,title --jq '[.[] | {number, title}]'` returns a JSON-array string as a tool result, the model processes it in the next turn without crashing. The exact pattern that was failing is now working.

Thanks for the quick fix.

PR #805, opened 2026-09-04
Preserve Qwen prefix cache across coding tool turns

darkolorin:
## Problem

Normal Qwen coding turns can discard the entire prefix even when the client echoes unchanged history. Two reproduced cases are a tool-only assistant echoed with `content: ""`, and a generated tool argument whose compact JSON, whitespace, or escaping changes when Hanashi parses and re-renders it. The latter passes the HTTP history comparison but fails Nagare's rendered-text prefix comparison. Re-encoding can also replace sampled token IDs in the encoding ledger while the backend still holds the original IDs.

## Change

- Omit exactly empty assistant text at the HTTP boundary; preserve whitespace, reasoning, other roles, and empty tool results. Log metadata-only reasons when the HTTP history gate resets the cache.
- Add a continuation policy for the bundled `HanashiConfig::Qwen35` encoding, which the tested Qwen3.6-27B package uses. Certify a fully closed assistant boundary only after backend stream teardown. Retain the actual sampled token ledger, parser, and decoder; use the canonical rendering of the authoritative history only to locate the new prompt suffix.
- Require an exact logical-history extension and an unchanged canonical prefix before appending. Invalidate reuse while preparing/generating, and recreate coherent encoding/backend state after incomplete output, cancellation, errors, or genuine edits. Other encodings retain their existing path.

No dependency, serialized configuration, binding, or GPU state-layout change. No rewind or partial truncation of Qwen's recurrent state.

## Validation

- **104 unit tests passed:** 72 CLI, 15 Hanashi, 17 Nagare. New regressions fail on the old behavior and cover empty assistant content, compact JSON, noncanonical sampled IDs, split Unicode, grouped tool results, changed history, incomplete tools, limits, cancellation, receiver drop, and preparation/state-creation failures.
- **Real Metal equivalence test passed** with local Qwen3.6-27B-M on an M3 Ultra: cached and fresh inference emitted identical greedy tokens through the stop token when given the same actual token transcript. Both short and long cases assert backend token identity after stream teardown and exact prefill counts. The test is opt-in via `UZU_CACHE_TEST_MODEL`.
- **31 HTTP requests passed** across blocking and streaming plain, tool, thinking, compact-JSON, empty-content, and interleaved conversations. Hits and expected cold/interleaved misses matched the assertions.
- **Both HTTP disconnect probes passed:** cancel a queued streaming or blocking request while another request is generating, then cancel the active request; verify correct recovery answers and exact cache accounting on the next continuation.
- **60 consecutive warm turns passed**, 20 each after 5,561-, 15,461-, and 31,961-token cold prompts. Every turn returned the required answer, completed its SSE stream, and reused exactly the previous prompt plus completion count. Median first-content latency was 318 / 367 / 452 ms; p95 was 321 / 370 / 455 ms.
- Release build, Rust formatting, formatter-discovery check, locked Cargo metadata, unchanged `Cargo.lock`, and diff whitespace checks passed. The tracked C/Metal formatting check passed for all 136 files using the installed LLVM formatter path.
- Dependency-policy validation (`cargo deny check -D warnings`), generated-file synchronization (`cli-tools sync --check`), and workspace Clippy (`cargo +nightly clippy --release --workspace --all-targets -- -D warnings`) passed. CI is pending on the final revision.

For the long GPU equivalence case, cached inference retained **5,999 tokens**, prefilled **22**, and returned its first token in **184 ms**; fresh inference on the identical 6,021-token input took **14,365 ms**. These are individual paired measurements, not benchmark averages. Separately, the original long HTTP JSON fixture now retains **5,912 tokens** and prefills **22** at **180 ms** TTFT; the earlier diagnostic baseline prefills 5,937 tokens at 14.07 s. The cold canonical prompt and preserved sampled transcript differ, so that HTTP comparison is not an identical-token oracle.

## Scope

This fixes the two reproduced false-miss paths. The server still retains one conversation, so an unrelated title/summary request can evict an active coding conversation; the interleaving probe confirms that limitation. The new continuation policy is deliberately limited to Qwen35, and incomplete/unsupported boundaries fall back to a full reset. The evidence does not identify a specific earlier PR as the introducing regression.

The GPU check exercised ordinary decoding (approximately one token per forward pass); active speculative decoding and other hardware/model packages still need their own validation.

### DIFF
--- a/crates/legacy/cli/src/server/chat_completions.rs
+++ b/crates/legacy/cli/src/server/chat_completions.rs
@@ -261,7 +261,9 @@ fn to_chat_messages(messages: &[OaiMessage]) -> Vec<ChatMessage> {
             if let Some(identifier) = &message.tool_call_id {
                 let result = tool_call_result_block(identifier, message.content.clone().unwrap_or_default());
                 chat_message = chat_message.with_block(result);
-            } else if let Some(content) = &message.content {
+            } else if let Some(content) = &message.content
+                && (chat_message.role != (ChatRole::Assistant {}) || !content.is_empty())
+            {
                 chat_message = chat_message.with_text(content.clone());
             }
             for tool_call in message.tool_calls.iter().flatten() {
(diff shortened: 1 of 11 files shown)


uuuvn:
Fixed in #806 with an order of magnitude less code

PR #811, opened 2026-09-07
Bound concurrent Metal build work

xhjkl:
This change limits concurrent Metal compilation to Cargo's configured job count. This keeps Metal compilation work across all source files within one shared limit.

Cold Metal generation: sampled peak **RSS 19.1 → 12.4 GiB, 35% lower**.
Elapsed time: 71.2 → 71.6 seconds.
Generated outputs identical; formatting, Metal check, and single-job rebuild passed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7674,6 +7674,7 @@ dependencies = [
  "half",
  "is_close",
  "itertools 0.14.0",
+ "jobserver",
  "keisoku",
  "libm",
  "monostate",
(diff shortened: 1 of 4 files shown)


uuuvn:
I'm not sure this is the right way to do it. https://doc.rust-lang.org/cargo/reference/build-scripts.html#jobserver

xhjkl:
@uuuvn, thanks for the pointer. Jobserver in Cargo was indeed something new to me — so I've reworked it that way.

PR #825, opened 2026-09-14
fix `syn::visit_mut` import

CC-Yeh:


uuuvn (approved):
Features should really be inside each crate, but fine given it's already like that

PR #830, opened 2026-09-15
Join 2+ in a row assistant messages

agolokoz:


uuuvn (changes requested):
Why this instead of allowing multiple assistant messages in a row?

PR #833, opened 2026-09-16
matmul: support trellis restart tapes

ry2009:
Stacked on #808. lalamo's `TrellisSpec` (trymirai/lalamo#364) fits a row as independent tapes of `restart_columns` weights, each with its own L-bit header, bit-concatenated; that layout beat the whole-row tape by 24 to 33% KL after QAT on our S packages, so it is what the packages carry. The kernels here walked one tape per row.

`TrellisConfig` gets `restart_columns: Option<u32>` and `TrellisParams` gets `tape_block_steps` and `tape_block_bits`; a whole-row tape is the one-block case and its addresses are unchanged. `trellis_decode.h`'s `Walk` now carries the lane's bit offset plus what one K group does to it: a constant when the group is whole tapes, a walk down the tape with a jump at its end otherwise. A tape is 136 bits at k=2 and 196 at k=3, so the word index and shift are recomputed once per K group instead of held loop-invariant. Restart widths are multiples of 16 that divide 512 or are multiples of it, so a GEMM K group (16 steps), a GEMV K block (32 or 128 steps) and a tape always nest.

Tests: two restart-64 tapes written by lalamo (k=2 and k=3) decode bit for bit by window read, by recurrence and to int8 levels; the oracle and bit-exact probe tests sweep the whole-row tape and restart 16, 64 (k=2 and k=3), 128 and 512, with probe points on both sides of every 64-column boundary; split-K partitions start mid-tape for restart 512.

Bench (pass of 2026-09-12, GEMV anchors drift -1.7%): restart-64 against the whole-row tape is within 0.3% at M 1 to 4 on every shape and 0 to 2.6% faster at M 6 and 8; the GEMM pass drifted 3.9% between anchors and is being rerun. `mlp_up_34816x5120` GEMV cells are unstable in both layouts (3 to 8 ms against 1.2 ms dense) and are not evidence either way.


ry2009:
Superseded by #864, which carries the S support (computed-codebook trellis kernels, restart-64 V4 and whole-row V2 leaves) onto current main in one PR.

PR #845, opened 2026-09-20
feat: Add Needle 3 as a shoji chat-via-message backend

XuQianJin-Stars:
## Summary
- Wrap Cactus Needle 3 (`libneedle` + `.cact`) as a `shoji` `chat_via_message` backend so on-device tool calling and JSON extraction work without changing `uzu-engine` kernels.
- Discover Needle from `NEEDLE3_LIB_PATH` / `NEEDLE_WEIGHTS_PATH` (or cache), register `cactus:needle3`, and skip quietly if the library or weights are missing.
- Expand `EngineConfig` with `allow_needle_usage` (default true) and optional Needle path fields; update Python / TypeScript / Swift `EngineConfig` surfaces.

```mermaid
flowchart LR
  App["App / nagare ChatSession"] --> Engine["uzu Engine"]
  Engine --> Registry["NeedleRegistry<br/>cactus:needle3"]
  Engine --> Backend["backend-needle"]
  Backend --> FFI["libloading<br/>libneedle.dylib"]
  FFI --> Weights["needle3.cact"]
  Backend --> Map["mapping<br/>tools / messages / envelope"]
  Map --> Output["shoji Output<br/>ToolCalls | Stop | Rejected"]
```

Needle is process-global (one loaded `.cact`, one C conversation). Multiple nagare sessions are simulated with a state owner plus reset/replay. Tool results are sent as a JSON array, matching Needle’s Python `run()`.

Not in this PR: Hugging Face auto-download, embeddings, iOS static linking, `ChatReplyStats.confidence`.

## Test plan
- [x] `cargo test -p backend-needle` (mapping + mock FFI, no real `libneedle`)
- [x] `UZU_NEEDLE_IT=1 cargo test -p uzu --test needle -- --ignored` with real Needle 3 lib + `.cact`
- [x] `Engine::new` still succeeds when Needle is absent or `allow_needle_usage = false`
- [x] Existing Mirai chat / tool-call examples unchanged
- [x] After Metal Toolchain is available: `cargo tools build swift` / python to regenerate UniFFI checksums


XuQianJin-Stars:
hi @alexxale could you take a look at this PR when you get a chance?

uuuvn:
This is very low quality and it looks like it just loads an actual engine binary from env var and spends 2.2KLoC just talking to it. If the model is good and its architecture fits nicely into our existing infrastructure in https://github.com/trymirai/lalamo and uzu-engine without a ton of custom code for it this might be interesting to have.

uuuvn:
Reading https://huggingface.co/Cactus-Compute/needle3

> Needle 3 is a Laddered Simple Attention Network, our small-model recipe: a Monarch Hadamard MLP in place of the FFN, GQA attention with causal conv taps, engram n-gram memory read by gather, and multi-lane hyper-connections, trained so that every depth from 2 to 20 layers is a deployable model. Most of its parameters sit in the engram, so the 121M model does the arithmetic of a 50M one. The weights are compressed to CQ2-bit with Cactus Quants; a byte-level grammar compiled from your schemas constrains every token, and every response carries a calibrated confidence score from a learned head.

Seems very complicated and a lot of things that are unique to this model so not worth the complexity to support properly in lalamo and engine. I'll close this.

